PR #13, opened 2020-03-23
Add Documentation For Configuration Templates

cmain:


docs/config_templates.md
```diff
@@ -0,0 +1,223 @@
+---
+title: 'Templates'
```
MichaelBaj: We should call this "Configuration Templates".  I have had numerous conversations over the weeks with customers and the term templates is used for a number of purposes (e.g. VMs with cloud-init installed so they can be rolled out more easily, and similarly for ISOs).

PR #22, opened 2020-04-14
AppID concept doc

ptimmons:
Includes update to sidebars.

### DIFF
--- a/docs/concepts_appid.md
+++ b/docs/concepts_appid.md
@@ -1,3 +1,8 @@
+---
+title: Application Identification
+sidebar_label: Application Identification
+---
+
 ## Identifying Applications by Name
 
 The 128T Session Smart routing platform has several techniques by which it can associate inbound traffic to named applications. This is sometimes referred to as "Application Identification" (or "appID"), and the feature that allows administrators to define network policies using Fully Qualified Domain Names (FQDNs) instead of IP addresses.
@@ -12,7 +17,7 @@ The 128T router has three built-in techniques for affiliating named applications
 
 Generally, when configuring a `service` on a 128T router, administrators use the `address` field to identify the IP address(es)/CIDR block(s) that the 128T should use to match to inbound traffic. However, the `address` field will also accept hostnames, such as `www.128technology.com`. When a service uses a hostname in the `address` field, it is referred to as a *DNS-based service*.
 
-For every DNS-based service, the 128T will use its local DNS to resolve the hostnames; for every IP address that is returned the 128T will treat that as though it were configured in the `address` field, and install FIB entries accordingly. Here is a sample configuration for reference:
+For every DNS-based service, the 128T will use its local DNS to resolve the hostnames; for every IP address that is returned the 128T will treat that as though it were configured in the `address` field, and install FIB entries accordingly. It will also refresh these FIB entries based on the time-to-live (TTL) returned by the DNS server. Here is a sample configuration for reference:
 
 ```
 admin@labsystem1.fiedler# show config running authority service interchange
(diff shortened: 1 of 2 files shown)


docs/concepts_appid.md
```diff
@@ -0,0 +1,236 @@
...
+            exit
+            share-service-routes  false
+        exit
+    exit
+exit
+```
+
+In this configuration, the `address` is configured as `community.128technology.com`. This will be resolved once the configuration is committed, and the 128T will resolve that hostname and install FIB entries. The 128T will periodically re-resolve those hostnames to ensure that the FIB entries are kept up-to-date.
+
+> Note: the names used in the `address` field must be DNS-resolvable names. You cannot use wildcards such as `*.128technology.com`.
```
MichaelBaj, line 48: unfortunately they don't render well in Typora, but Docusaurus uses something called [admonitions/callouts](https://v2.docusaurus.io/docs/markdown-features/#calloutsadmonitions).  I have been using them throughout the docs site.  In this case you would use `:::note` followed by text and a trailing `:::`.

docs/concepts_appid.md
```diff
@@ -0,0 +1,236 @@
...
+2. Unlike AppID based on TLS, there is no "chicken and egg issue;" i.e., no prerequisite to have the destination be reachable in order to learn the destination's name.
+
+**Disadvangates**:
+
+This technique relies on your 128T getting the same answers from DNS that your clients get. Take for example a massive SaaS provider such as `www.salesforce.com`. Their web site resolves to hundreds or thousands of IP addresses distributed across the globe. When the 128T resolves that hostname, it will get a handful of answers; if a client does the same DNS query and gets a different set of answers, then the traffic the client sends will not match the FIB entries that the 128T has created.
+
+*Key to Success*: Use DNS-based services when your destination uses a small set of nonvolatile IP addresses. This technique is perfect for small, hosted services. The effectiveness of using DNS-based services diminishes rapidly when referencing names that resolve to dozens of address or more.
```
MichaelBaj, line 59: Following the above comment, I have been treating the Key To Success (I like this) as a callout using the syntax ```:::tip Key to success`.  You can define your own heading using the styling of a different admonition.

docs/concepts_appid.md
```diff
@@ -0,0 +1,236 @@
...
+
+### AppID based on TLS
```
MichaelBaj, line 61: Note: The new APP ID functionality will be changing this behavior.  I believe you have been involved in the discussions.  For the purposes of the current version of software, this documentation is correct.

docs/concepts_appid.md
```diff
@@ -0,0 +1,236 @@
...
+
+            application-identification
+                mode  module
+            exit
+        exit
+    exit
+exit
+```
+
+Next, scripts are placed on the router's filesystem at `/etc/128technology/application-modules/`. These scripts must produce JSON output stored at `/var/run/128technology/application-modules/`, which in turn will be consumed by the 128T router and installed as FIB entries.
```
MichaelBaj, line 181: you may want to add that while not as dynamic as using a script, you can also just add a static JSON file to the `application-modules` directory to define service addresses.

docs/concepts_appid.md
```diff
@@ -0,0 +1,236 @@
...
+        "ip-prefix": "3.22.11.0/24"
+      },
+      ...
+    ]
+  },
+  "module-name": "zoom"
+}
+```
+
+> Note: in the case of this module, the script that generates the JSON is `/etc/128technology/application-modules/zoom.py`, the output is stored as `/var/run/128technology/application-modules/zoom.json`
```
MichaelBaj, line 232: would you consider adding the contents of `zoom.py` as an appendix to this guide?

docs/concepts_appid.md
ptimmons: Nice, will do.

docs/concepts_appid.md
ptimmons: Yeah, I'm just documenting what's there now with the understanding that it'll need to be rewritten for new behavior.

docs/concepts_appid.md
ptimmons: office365 includes a python executable too.

ptimmons: office365 still relies on a python module too.

PR #23, opened 2020-04-15
Update the monitoring agent documentation in preparation for the latest release

agrawalkaushik:
The new versions of monitoring agent 1.1.0 and 2.0.0 add some new features and functionality to the product. The list of some of the enhancements as as follows:

* Add a new top analytics collector for capturing the top sessions, sources and applications
* Create a new test command for verify input configuration and its results
* Update to the latest version of stable telegraf release (v 1.14.0)
* Automatically stage input configuration for all 128T developed collectors along with the default settings.

### DIFF
--- a/docs/plugin_monitoring_agent.md
+++ b/docs/plugin_monitoring_agent.md
@@ -18,55 +18,58 @@ The monitoring agent at its core is designed to be able to push data to external
 
 ## Installation
 
-The 128T Monitoring Agent can be obtained from the official 128T software repository and is installed using the dnf utility:
-```
-dnf install 128T-monitoring-agent
-```
+The 128T Monitoring Agent can be obtained from the official 128T software repository. The following versions of the **monitoring-agent** are available for corresponding 128T software version.
 
-example:
+| Monitoring Agent | 128T |
+| --- | --- |
+| 128T-monitoring-agent-1.1.0 | 128T >= 4.1.0; 128T < 4.3.0 |
+| 128T-monitoring-agent-2.0.0 | 128T >= 4.3.0 |
 
+The agent can be install using dnf utility. For example.
 ```
 # dnf install 128T-monitoring-agent
-128 Technology 7 - x86_64                                                                    13 MB/s |  30 MB     00:02
+Last metadata expiration check: 0:04:59 ago on Wed 15 Apr 2020 06:25:48 AM UTC.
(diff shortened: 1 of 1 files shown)


docs/plugin_monitoring_agent.md
```diff
@@ -18,55 +18,58 @@ The monitoring agent at its core is designed to be able to push data to external
...
-The 128T Monitoring Agent can be obtained from the official 128T software repository and is installed using the dnf utility:
-```
-dnf install 128T-monitoring-agent
-```
+The 128T Monitoring Agent can be obtained from the official 128T software repository. The following versions of the **monitoring-agent** are available for corresponding 128T software version.
 
-example:
+| Monitoring Agent | 128T |
+| --- | --- |
+| 128T-monitoring-agent-1.1.0 | 128T >= 4.1.0; 128T < 4.3.0 |
```
MichaelBaj, line 25: When the monitoring agent becomes a plugin, will this go away?  Meaning the administrator shouldn't have to worry about installing the right version.  We should take care of that for them through the plugin portal.

agrawalkaushik: We can definetly automate the selection through a salt state. The plugin will check for the running software version of 128T on the router and install the correct version.

agrawalkaushik:
> Looks good, just wondering about whether specifying the version is even necessary. Is Top Analytics Collector unique to version 2.0.0? If so, then the ## History section is necessary specifying that this new capability has been added to this version of plugin.

The new collector is backward compatible. However, in 4.3 we broke the peer-path metrics by changing the parameters we use for querying those metrics. This was forcing folks to have to modify the pre-canned default-config file for metrics. The two versions, are unfortunately, needed because of that breaking change.

docs/plugin_monitoring_agent.md
MichaelBaj: If you were to `dnf install 128T-monitoring-agent` due to pinning, isn't the correct version going to be automatically installed?

agrawalkaushik: so the problem is that if you just do `dnf install 128T-monitoring-agent` it will try to upgrade the 128T version along with it.

MichaelBaj: That is very unfortunate. In that case, a plugin is the best path forward.

PR #26, opened 2020-04-15
Pcli command reference 1

MichaelBaj:


docs/cli_reference.md
```diff
@@ -7176,13 +7176,53 @@ Available to _admin_ only.
...
+
+##### Keyword Arguments
+- **force**  Skip confirmation prompt
+- **node**   Node for which to display transform status
+- **router** Router for which to display transform status
+
+#### Description
+Display the status of UDP transform between peers.
+
+UDP transform, also known as the [firewall detector]](concepts_machine_communication.md#firewall-detector) is used to determine if stateful firewalls exist between 128T peers as certain firewalls may interfere with SVR.
```
ptimmons, line 7196: A 128T router may need to transform TCP packets into UDP packets to enable SVR to traverse stateful firewalls. By default, the 128T router runs a [firewall detector]](concepts_machine_communication.md#firewall-detector) process over peer paths, and will dynamically enable UDP transform when necessary. (Administrators may also elect to enable UDP transform if they know there are stateful firewalls in the path.) This command shows whether a path has UDP transform enabled, and if so, which firewall detection tests triggered the feature to be enabled.

MichaelBaj: Thank you, that reads much better. What did you think about the other content/formatting. i.e. 
##### Keyword Arguments
##### Positional Arguments
##### See Also

Also considering changing #### Syntax to #### Usage to be consistent with the PCLI output.  Should there also be an #### Example section?

ptimmons: Yeah, I like it much better. Good idea on the #### Example section too.

PR #31, opened 2020-04-22
BCP monitoring head ends

ptimmons:
First version.


ptimmons:
> Would you consider adding a link to [The Monitoring Agent](plugin_monitoring_agent.md) to point people to using a tool of our creation to start their monitoring journey down a path we prefer?

Yeah, what I would love to (eventually) include is a table that has the various means for data acquisition: SNMP, REST, MonAgent, etc., along with a table that checks off which metrics are supported through which means. Also, a sample MonAgent config fragment for grabbing the data we recommend.

But this was just in the interest of putting something out there, even if it's not the ideal end state.

MichaelBaj:
Created task DOCS-81 to track updating this monitoring document with configuration references.

PR #37, opened 2020-04-28
Add release notes for monitoring agent

agrawalkaushik:
Fix a couple of bugs affecting 1.1.0 and 2.0.0 version of the monitoring-agent.
Updated the release notes with the data on the issues fixed

DOCS-86 #time 20m

### DIFF
--- a/docs/plugin_monitoring_agent.md
+++ b/docs/plugin_monitoring_agent.md
@@ -22,19 +22,19 @@ The 128T Monitoring Agent can be obtained from the official 128T software reposi
 
 | Monitoring Agent | 128T |
 | --- | --- |
-| 128T-monitoring-agent-1.1.0 | 128T >= 4.1.0; 128T < 4.3.0 |
-| 128T-monitoring-agent-2.0.0 | 128T >= 4.3.0 |
+| 128T-monitoring-agent-1.1.1 | 128T >= 4.1.0; 128T < 4.3.0 |
+| 128T-monitoring-agent-2.0.1 | 128T >= 4.3.0 |
 
 The agent can be install using dnf utility. For example.
 ```
 # dnf install 128T-monitoring-agent
-Last metadata expiration check: 0:04:59 ago on Wed 15 Apr 2020 06:25:48 AM UTC.
+Last metadata expiration check: 0:00:31 ago on Tue 28 Apr 2020 02:40:10 PM UTC.
 Dependencies resolved.
 ==============================================================================================================================================================================================================================================================
  Package                                                               Arch                                                   Version                                                   Repository                                                       Size
 ==============================================================================================================================================================================================================================================================
 Installing:
- 128T-monitoring-agent                                                 x86_64                                                 2.0.0-3                                                   plugins-staging                                                 5.0 M
+ 128T-monitoring-agent                                                 x86_64                                                 2.0.1-2                                                   plugins-staging                                                 5.0 M
 Installing dependencies:
  telegraf-128tech                                                      x86_64                                                 1.14.0-1                                                  128tech                                                          20 M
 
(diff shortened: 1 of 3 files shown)


docs/release_notes_monitoring_agent_1.1.0.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: 128T Monitoring Agent 1.1 Release Notes
-sidebar_label: 1.1
+title: 128T Monitoring Agent 1.1.0 Release Notes
+sidebar_label: 1.1.0
```
MichaelBaj, line 3: Some numerical values ending in 0 sometimes need to be quoted (e.g. '1.0').

sidebars.js
```diff
@@ -127,8 +127,10 @@ module.exports = {
         "type": "category",
         "label": "128T Monitoring Agent",
         "items": [
-          "release_notes_monitoring_agent_2.0",
-          "release_notes_monitoring_agent_1.1",
+          "release_notes_monitoring_agent_2.0.1",
```
MichaelBaj, line 130: I only create new files for major and minor versions of the release notes.  Because patches are built on top of the minor release, it makes sense for all the patches to be contained within a single document.  See the 128T release notes.  This is a convention followed by other software products as well (not to mention that it reduces the number of files to manage).

sidebars.js
```diff
@@ -127,8 +127,10 @@ module.exports = {
...
+          "release_notes_monitoring_agent_2.0.0",
+          "release_notes_monitoring_agent_1.1.1",
+          "release_notes_monitoring_agent_1.1.0",
```
MichaelBaj, line 133: Please collapse these release notes into two separate files.  One for all 2.0 and 1.1 patches.

sidebars.js
agrawalkaushik: Ah, that makes sense. I'll just add a section in the major version file then.

docs/release_notes_monitoring_agent_1.1.1.md
```diff
@@ -0,0 +1,21 @@
...
+### Issues Fixed
+
+----
+- **MON-185** telegraf error when processing results from peer path input
+
+  _**Resolution:**_ The extra logging causing the problem was removed
+------
+- **MON-186** LTE metric collector not reporting any values
+
+  _**Resolution:**_ Fix the python imports causing the data to be missing
```
jeffmart, line 16: typo, Fix I think should be "Fixed"

docs/release_notes_monitoring_agent_2.0.1.md
```diff
@@ -0,0 +1,21 @@
...
+### Issues Fixed
+
+----
+- **MON-185** telegraf error when processing results from peer path input
+
+  _**Resolution:**_ The extra logging causing the problem was removed
+------
+- **MON-186** LTE metric collector not reporting any values
+
+  _**Resolution:**_ Fix the python imports causing the data to be missing
```
jeffmart, line 16: Same here "Fixed"

sidebars.js
```diff
@@ -127,8 +127,10 @@ module.exports = {
...
         ],
       },
```
jeffmart, line 135: I like the "_**Resolution:**_" part, I do think that we can come up with better canned references for future release notes. For example, instead of:

"Fixed the syntax in the sample and staged configuration for events input"

Could be:

"Updated the sample and staged configuration example for events"

Or:

"Fix the python imports causing the data to be missing"

Could be:

"Updated the library imports to display the missing data"

Not saying we need to do this now, just putting the comment out there

agrawalkaushik: I have to make some edits anyways, I'll include these changes.

PR #45, opened 2020-05-08
Add Using Saltstack at Scale with 128T doc

tyc1031:
Adds a document to the best practices section which describes
 the best practices for using saltstack at scale with the 128T
 conductor.

### DIFF
--- a/docs/bcp_salt.md
+++ b/docs/bcp_salt.md
@@ -1,0 +1,167 @@
+---
+title: Using Saltstack at Scale With 128T
+sidebar_label: Saltstack at Scale With 128T
+---
+
+## Abstract
+
+This guide is intended to provide methods for users to implement custom Saltstack functionality in a performant way that will not affect their ability to provision new systems at scale.
+
+This guide is not intended to teach the users how to use Saltstack. This document will reference the Salt top file and Salt pillars. Some good resources for these topics are:
+
+[Salt Top File](https://docs.saltstack.com/en/latest/ref/states/top.html)
+
+[Salt Pillars](https://docs.saltstack.com/en/latest/topics/tutorials/pillar.html)
+
+## Using Salt with 128T
+
+The 128T Conductor uses different Salt environments to logically separate the 128T specific Salt state and modules from any custom Salt modules implemented by the user. The locations for each environment are defined in the Salt Master configuration file on the Conductor located at `/etc/128technology/salt/master`:
+```
+file_roots:
+  128T:
+    - /usr/lib/128technology/python/salt/file_roots
+  base:
+    - /srv/salt
+  plugins:
+    - /var/lib/128technology/plugins/salt
+
+pillar_roots:
+  128T:
+    - /usr/lib/128technology/python/salt/pillar
+  base:
+    - /srv/pillar
+  plugins:
+    - /var/lib/128technology/plugins/pillar
+```
+
+
+The `128T` and `plugins` environments are managed by the Conductor and should not be touched. The `base` environment is left for the user to implement any custom Salt logic for their specific deployment. When a Salt Minion connects to the 128T Conductor the Salt Master will execute the Salt highstate for each Salt environment automatically. The highstate consists of the states listed in the `top.sls` for each environment. By default the `top.sls` for the `base` environment performs a dummy state meant to serve as an example to users:
+```
+[root@t101-dut1 ~]# cat /srv/salt/top.sls
+base:
+  '*':
+    - dummy
+[root@t101-dut1 ~]# cat /srv/salt/dummy.sls
+/dev/null:
+  file.touch:
+    - name: /dev/null
+```
+
+## Using Salt Pillars
+
+Pillars are tree-like structures of data defined on the Salt Master and passed through to Salt Minions. They allow confidential, targeted data to be securely sent only to the relevant Salt Minion. Pillars are commonly used to store variable information for each Salt Minion, which are used when the Salt Minion renders a Salt state before executing it. Here is an example of a common approach for using Salt pillars:
+
+The pillar top file defines which pillars apply to which Salt Minions (Note: The ellipsis is not part of the syntax, it serves to illustrate there are a large number of Salt Minions):
+
+`/srv/pillar/top.sls`:
+```
+base:
+  '*':
+    - common
+  'Router1NodeA':
+    - router1A
+  'Router1NodeB':
+    - router1B
+
+   ...
+
+  'Router999NodeA':
+    - router999A
+  'Router999NodeB':
+    - router999B
+```
+
+Each pillar file defined above contains the variables for a specific Salt Minion:
+
+`/srv/pillar/router1A.sls`:
+```
+configured_interfaces:
+  - name: eth0
+    address: '192.168.1.1'
+    netmask: '255.255.255.254'
+    gateway: '192.168.1.2
+  - name: eth1
+    address: '172.16.0.1'
+    netmask: '255.255.255.0'
+    gateway: '172.16.0.100'
+
+```
+
+`/srv/pillar/router1B.sls`:
+```
+configured_interfaces:
+  - name: eth0
+    address: '192.168.1.2'
+    netmask: '255.255.255.254'
+    gateway: '192.168.1.1'
+  - name: eth1
+    address: '172.16.0.2'
+    netmask: '255.255.255.0'
+    gateway: '172.16.0.100'
+```
+
+The pillar data can be accessed as variables in Salt states and the Salt Minion will automatically render the state with the correct pillar information when the state is executed:
+```
+{%- set configured_interfaces = pillar.get('configured_interfaces') %}
+{%- for interface in configured_interfaces %}
+
+Interface {{ interface.name }}:
+  network.managed:
+    - enabled: True
+    - type: eth
+    - ipaddr: {{ interface.address }}
+    - netmask: {{ interface.address }}
+    - gateway: {{ interface.gateway }}
+    - dns:
+      - 8.8.8.8
+      - 8.8.4.4
+```
+
+This approach quickly breaks down at scale. Each time a highstate is run the Salt Master needs to parse the pillar top file and decide which pillar files apply to each Salt Minion. The top file supports glob matching and is not always a simple 1:1 matching from Salt Minion ID to pillar file, therefore the entire file needs to be parsed each time a highstate is performed. Next the Salt Master encrypts the pillar data and sends it to each Salt Minion. These operations become extremely costly and bring the Salt Master to a halt if the pillar top file contains 1000s of individual pillar files.
+
+## A Better Approach: Map Files
+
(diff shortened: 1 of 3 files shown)


docs/bcp_salt.md
```diff
@@ -0,0 +1,167 @@
...
+  128T:
+    - /usr/lib/128technology/python/salt/pillar
+  base:
+    - /srv/pillar
+  plugins:
+    - /var/lib/128technology/plugins/pillar
+```
+
+
+The `128T` and `plugins` environments are managed by the Conductor and should not be touched. The `base` environment is left for the user to implement any custom Salt logic for their specific deployment. When a Salt Minion connects to the 128T Conductor the Salt Master will execute the Salt highstate for each Salt environment automatically. The highstate consists of the states listed in the `top.sls` for each environment. By default the `top.sls` for the `base` environment performs a dummy state meant to serve as an example to users:
```
MichaelBaj, line 38: "... should not be modified" (instead of touched).

docs/bcp_salt.md
```diff
@@ -0,0 +1,167 @@
+---
+title: Using Saltstack at Scale With 128T
+sidebar_label: Saltstack at Scale With 128T
+---
+
+## Abstract
```
MichaelBaj, line 6: "Introduction"

docs/bcp_salt.md
```diff
@@ -0,0 +1,167 @@
...
+    - dummy
+[root@t101-dut1 ~]# cat /srv/salt/dummy.sls
+/dev/null:
+  file.touch:
+    - name: /dev/null
+```
+
+## Using Salt Pillars
+
+Pillars are tree-like structures of data defined on the Salt Master and passed through to Salt Minions. They allow confidential, targeted data to be securely sent only to the relevant Salt Minion. Pillars are commonly used to store variable information for each Salt Minion, which are used when the Salt Minion renders a Salt state before executing it. Here is an example of a common approach for using Salt pillars:
```
MichaelBaj, line 52: "They allow confidential..."
Is it more accurate to say, something to the effect of "Salt pillars are intended to be used for confidential information as the information is encrypted and sent securely to the respective Salt Minion.

docs/bcp_salt.md
```diff
@@ -0,0 +1,167 @@
...
+```
+
+
+The `128T` and `plugins` environments are managed by the Conductor and should not be touched. The `base` environment is left for the user to implement any custom Salt logic for their specific deployment. When a Salt Minion connects to the 128T Conductor the Salt Master will execute the Salt highstate for each Salt environment automatically. The highstate consists of the states listed in the `top.sls` for each environment. By default the `top.sls` for the `base` environment performs a dummy state meant to serve as an example to users:
+```
+[root@t101-dut1 ~]# cat /srv/salt/top.sls
+base:
+  '*':
+    - dummy
+[root@t101-dut1 ~]# cat /srv/salt/dummy.sls
```
MichaelBaj, line 44: Be consistent within the file.  Example below use  single back ticks for the command and triple back ticks for content.

docs/bcp_salt.md
```diff
@@ -0,0 +1,167 @@
...
+    - type: eth
+    - ipaddr: {{ interface.address }}
+    - netmask: {{ interface.address }}
+    - gateway: {{ interface.gateway }}
+    - dns:
+      - 8.8.8.8
+      - 8.8.4.4
+```
+
+This approach quickly breaks down at scale. Each time a highstate is run the Salt Master needs to parse the pillar top file and decide which pillar files apply to each Salt Minion. The top file supports glob matching and is not always a simple 1:1 matching from Salt Minion ID to pillar file, therefore the entire file needs to be parsed each time a highstate is performed. Next the Salt Master encrypts the pillar data and sends it to each Salt Minion. These operations become extremely costly and bring the Salt Master to a halt if the pillar top file contains 1000s of individual pillar files.
```
MichaelBaj, line 120: It is valuable to quantify any scaling limits.  Obviously this is heavily dependent on the hardware platform, however we should provide some kind of guidance. Remember people will read and follow these guides.  Best to be conservative and suggest a point where the admin should understand when this is relevant to them.

Instead of "bring to a halt", cause the Salt Master to be unable to process any minion requests. - try to be more specific.

docs/bcp_salt.md
```diff
@@ -0,0 +1,167 @@
...
+
+## A Better Approach: Map Files
+
+Pillar files are best used only for sensitive data like passwords or SSH keys. Any other data should be converted to use map files instead. The swap from pillar files to map files is very simple.
+
+Move all pillar files from the pillar directory `/srv/pillar/` to a new data directory within the Salt file roots `/srv/salt/data` and rename the pillar files to have the extension matching their data type instead of the `.sls` extension:
+```
+[root@t101-dut1 ~]# ll /srv/salt/data/
```
MichaelBaj, line 128: follow the convention above.  Separate line for the command from the output

docs/bcp_salt.md
```diff
@@ -0,0 +1,167 @@
...
+      - 8.8.4.4
+```
+
+This approach quickly breaks down at scale. Each time a highstate is run the Salt Master needs to parse the pillar top file and decide which pillar files apply to each Salt Minion. The top file supports glob matching and is not always a simple 1:1 matching from Salt Minion ID to pillar file, therefore the entire file needs to be parsed each time a highstate is performed. Next the Salt Master encrypts the pillar data and sends it to each Salt Minion. These operations become extremely costly and bring the Salt Master to a halt if the pillar top file contains 1000s of individual pillar files.
+
+## A Better Approach: Map Files
+
+Pillar files are best used only for sensitive data like passwords or SSH keys. Any other data should be converted to use map files instead. The swap from pillar files to map files is very simple.
+
+Move all pillar files from the pillar directory `/srv/pillar/` to a new data directory within the Salt file roots `/srv/salt/data` and rename the pillar files to have the extension matching their data type instead of the `.sls` extension:
```
MichaelBaj, line 126: Does the extension have to be `.yaml`?  If so, I would be explicit about needing to replace .sls with .yaml.

docs/bcp_salt.md
```diff
@@ -0,0 +1,167 @@
...
+Now convert all Salt states to access the information from the map file instead of the pillar file.
+
+Before:
+```
+{%- set configured_interfaces = pillar.get('configured_interfaces') %}
+```
+
+After:
+
+*Note*: In this example the map files for each Salt Minion are named after the Salt Minion ID. The first line uses the Salt Minion ID to form the file name.
```
MichaelBaj, line 148: Use an admonition. e.g.
:::note
some note here
:::

docs/bcp_salt.md
```diff
@@ -0,0 +1,167 @@
...
+{%- import_yaml data_file as data %}
+{%- set configured_interfaces = data.get('configured_interfaces') %}
+```
+
+There is no need to manually sync the map file from the data directory on the Salt Master to the Salt Minions. Since the data files located at `/srv/salt/data` are placed within the Salt file roots `/srv/salt/` the Salt Minion can fetch them from the Salt Master automatically. Within the states the data file path is referenced as `data/` as seen in the first line of the example above because the Salt state's root directory is the Salt file roots.
+
+
+## Conclusion
+
+With the Salt pillar approach the Salt Master renders the entire pillar top file and encrypts the pillar data each time it needs to perform highstate. With the map file approach the Salt Master simply executes the highstate and the Salt Minion will fetch the correct map file automatically and render the information locally, saving lots of CPU cycles on the Salt Master. The only downside with the map approach is that the data is not encrypted on the Salt Minion when the map file retrieved and cached locally, which is why the map approach should not be used for sensitive data.
```
MichaelBaj, line 160: Is the map file data still sent over the encrypted SSH tunnel between minion and master?

sidebars.js
```diff
@@ -93,6 +93,7 @@ module.exports = {
       "bcp_sdwan_design_guide",
       "bcp_service_and_service_policy_design",
       "bcp_tenants",
+      "bcp_salt",
```
MichaelBaj, line 96: I can envision us having more best practice guides around the use of salt.  Consider renaming the file to `bcp_salt_pillars.md`

docs/bcp_salt.md
```diff
@@ -0,0 +1,167 @@
...
+base:
+  '*':
+    - dummy
+[root@t101-dut1 ~]# cat /srv/salt/dummy.sls
+/dev/null:
+  file.touch:
+    - name: /dev/null
+```
+
+## Using Salt Pillars
```
jamesesilvia, line 50: While these are good description of common salt/pillar example it's a lot of information and you include pointers to salt/pillar documentations. This section could be stripped down a bit while making it obvious this is bad for scale with bolidng or something. I just want the `A Better Approach` to pop at the top but this is just my opinion

docs/bcp_salt.md
ptimmons, line 120: s/1000s/thousands

sidebars.js
tyc1031: done

docs/bcp_salt.md
tyc1031: done

docs/bcp_salt.md
tyc1031: done

docs/bcp_salt.md
tyc1031: I went back and forth on how much information I should include. Ultimately when I read the original map file blog post I felt it did not have enough information even for an experienced salt user. I ended up including all these examples because I wanted a junior team member with some salt experience to be able to easily get map file up and running for a customer deployment. I feel that the map file section is short and easy to understand only because I explain how pillars are deployed first, so I can easily illustrate if your deployment looks like this and how to fix it. I don't wouldn't be able to start the map files sections with "copy these files from A to B" if i I give some context about why there are files a location A in the first place. If other people have opinions on the amount of information here please chime in.

docs/bcp_salt.md
tyc1031: Fixed and reworded the rest of the paragraph

docs/bcp_salt.md
tyc1031: fixed

docs/bcp_salt.md
tyc1031: fixed

docs/bcp_salt.md
tyc1031: added more specific information and symptoms

docs/bcp_salt.md
tyc1031: Nope, clarified json can be used as well

docs/bcp_salt.md
tyc1031: ooo nice that looks much better, fixed

docs/bcp_salt.md
tyc1031: yes, added that clarification

PR #54, opened 2020-05-19
Monitoring Agent 1.2.0 and 2.1.0 updates

gregschrock:
## Description

Update the monitoring agent documentation based on the latest changes. Updates include:

- apply some linting
- replace version description
- document the new `stop` command
- update the events sample
- add the arp state collector


docs/plugin_monitoring_agent.md
```diff
@@ -7,69 +7,72 @@ The 128T Monitoring Agent is an entity for collecting data from a node running 1
...
 
 The 128T Monitoring Agent can be obtained from the official 128T software repository. The following versions of the **monitoring-agent** are available for corresponding 128T software version.
 
-| Monitoring Agent | 128T |
-| --- | --- |
-| 128T-monitoring-agent-1.1.1 | 128T >= 4.1.0; 128T < 4.3.0 |
-| 128T-monitoring-agent-2.0.1 | 128T >= 4.3.0 |
+| Monitoring Agent            | 128T                        |
+| --------------------------- | --------------------------- |
+| 128T-monitoring-agent-1.2.0 | 128T >= 4.1.0; 128T < 4.3.0 |
```
MichaelBaj, line 25: Is there a new version of the monitoring agent?  I am presuming the modifications to this document are indicative of the new version.  If so, this PR should be inclusive of the release notes for the new version so they are released at the same time.

docs/plugin_monitoring_agent.md
```diff
@@ -78,31 +81,31 @@ Complete!
...
 
-`sample-interval` and `push-interval` indicate the frequency (in seconds) for how often the data is collected and subsequently pushed to the collection target. When the `push-interval` value is greater than the `sample-interval`, the agent will produce `ceiling(push-interval/sample-interval)` samples collected within the push duration.  It is recommended to configure the `push-interval` as a multiple of `sample-interval`.
+`sample-interval` and `push-interval` indicate the frequency (in seconds) for how often the data is collected and subsequently pushed to the collection target. When the `push-interval` value is greater than the `sample-interval`, the agent will produce `ceiling(push-interval/sample-interval)` samples collected within the push duration. It is recommended to configure the `push-interval` as a multiple of `sample-interval`.
```
BenMatase, line 109: I was considering saying that we should add a section explaining the `lib-directory` option, but I realized that the end user shouldn't need to use it.

docs/plugin_monitoring_agent.md
```diff
@@ -422,25 +436,33 @@ The `name` becomes the name of the measurement in the context of influxdb format
...
+  # topic = "events"
 
   ## input event filtering based on type (admin, alarm, system, traffic, provisioning)
   ## NOTE: For information on filtering severity refer to the output configuration example
-  [inputs.execd.tagpass]
+  [inputs.t128_events.tagpass]
   type = ["alarm", "admin"]
 ```
 
 ### Device interface state collector
```
BenMatase, line 461: NBD, but maybe make all of these capitalized to be consistent with the other titles

docs/plugin_monitoring_agent.md
gregschrock: Yes, Kaushik was going to put up the release notes. I'm imagining them being added on top of the current changes here or being a separate PR that is merged at the same time.

How is the timing between docs updates and the actually RPM publishing usually handled?

MichaelBaj: The release notes should be included in the same PR along with any related documentation updates.

The process is:
1. post PR, approve documentation.  Do not merge
2. publish software assets.
3. merge documentation.

technically you can have multiple PRs, you just need to make sure they are all merged at the same time.

docs/plugin_monitoring_agent.md
MichaelBaj: Ben is correct.  Convention for headings is to have all words be capitalized.

docs/plugin_monitoring_agent.md
gregschrock: Yeah, ATM it's mostly for internal use. But I do think you're right. It should be documented here.

docs/plugin_monitoring_agent.md
jeffmart: Also the person performing the "publish software assets" should also be doing the merge documentation as part of the release process.

docs/plugin_monitoring_agent.md
```diff
@@ -422,25 +436,33 @@ The `name` becomes the name of the measurement in the context of influxdb format
 
 The event collector can be used for collecting and pushing events for various categories such as admin, alarm, system, traffic and provisioning as they occur on the system. The type of the event is available via a `tag` and can be used for filtering only specific events as desired. For example, the following configuration can be used for pushing just the `alarm` and `admin` event
 
-```
-[[inputs.execd]]
+```toml
+[[inputs.t128_events]]
```
jeffmart, line 440: If this change is because of the newer version (as I expect it is) keeping the older supported options in the documentation may be needed. If they are deprecated they should be identified as being deprecated and a reference added to the release notes as such.

gregschrock: Old versions of the config using `execd` will function as they did before. The executable still exists and has the same default behavior, so everything is backward compatible. We have made some additions and are updating the sample to reflect that additional functionality.

docs/release_notes_monitoring_agent_1.X.md
```diff
@@ -1,7 +1,49 @@
...
+
+---
+- **MON-184** Added stop command in cli to stop all associated Telegraf services.
+
+For help using this cli option, please refer to the [Monitoring Agent Guide](plugin_monitoring_agent.md#stopping-services).
+
+---
+- **MON-141** Added support for multiple logically seperate monitoring agent instances with the `lib-directory` config option.
+
+For help configuring this option, please refer to the [Monitoring Agent Guide](plugin_monitoring_agent.md#configuration).
```
gregschrock, line 17: Does this configuration link work? My linter was complaining about multiple `Configuration` headings as they cause link issues.

gregschrock, line 17: Does this configuration link work? My linter was complaining about multiple Configuration headings as they cause link issues.

BenMatase: It works when I spin it up. Maybe it is just grabbing the first one which just so happens to be the correct one. Maybe I could change the second one to be `Configure` and all of those cli associated headers to be verbs?

MichaelBaj: There are no limits to the number of references to a single or multiple documents.  Greg, do you have a link to the run that failed for you?

gregschrock: Yeah Ben, I thought about rewording those CLI headers. I'm not sure it's worth it, but maybe.

MichaelBaj: The "trick" that I do is once the page has been rendered, you can click on the hash mark that appears next to the header.  Copy that URL (don't forget to add the `.md` to the file name) and add that to your markdown.  I know that multiple headers of the same name get different internal links.  I believe they get a trailing 1, 2, 3, etc.  You have to look at the HTML.

sidebars.js
```diff
@@ -139,8 +139,8 @@ module.exports = {
         "type": "category",
         "label": "128T Monitoring Agent",
         "items": [
-          "release_notes_monitoring_agent_2.0",
-          "release_notes_monitoring_agent_1.1",
+          "release_notes_monitoring_agent_2.X",
+          "release_notes_monitoring_agent_1.X",
```
MichaelBaj, line 143: Does the monitoring agent follow a major.minor versioning scheme?  Do you intend to release patches for minors?  If it is going to follow a similar scheme as the 128T, then the file names, title, sidebar labels should follow suit.  Please look at those release notes as a reference.

agrawalkaushik: The monitoring agent does follow the major.minor version scheme. So far we had to release a patch with bug fixes only for 1.1 and 2.0 versions. I agree with @MichaelBaj that we should be consistent with our other software releases. I think we should add a `1.2` and `2.1` to the side bar instead.

docs/plugin_monitoring_agent.md
```diff
@@ -7,110 +7,126 @@ The 128T Monitoring Agent is an entity for collecting data from a node running 1
...
+  - name: file
+  - name: message_queue
 ```
 
-The `enabled` field is meant as global toggle for applying the monitoring agent functionality.  When set to `disabled` the monitoring agent will remain dormant on the 128T.
+The `enabled` field is meant as global toggle for applying the monitoring agent functionality. When set to `disabled` the monitoring agent will remain dormant on the 128T.
+
+`lib-directory` is the root directory for the Monitoring Agent. Other directores exist relative to it. This is useful when intending to isolate a set of Monitoring Agent elements from others.
+
+:::caution
```
agrawalkaushik, line 115: If we fully support this feature then we shouldn't caution people from not using it. I realize we did this change to support some internal use cases but I feel this caution is not warranted.

docs/plugin_monitoring_agent.md
```diff
@@ -281,22 +297,20 @@ This example sends data to a kafka broker:
...
-```
+```toml
 [[outputs.syslog]]
   address = "udp://<ip>:514"
   default_sdid = "128T"
 ```
 
-:::important
-For syslog output, not specifying the `default_sdid` parameter can result in empty or truncated messages
+:::important
```
agrawalkaushik, line 312: Nit: We should try to avoid having trailing white spaces wherever possible.

docs/plugin_monitoring_agent.md
```diff
@@ -436,11 +469,37 @@ The event collector can be used for collecting and pushing events for various ca
   type = ["alarm", "admin"]
 ```
 
-### Device interface state collector
+:::important  
+In versions 1.2.0, 2.1.0 and later, the `execd` input should be replaced by the more feature rich `t128_events` input type described below.  
+:::
 
-The `deviceInterfaceStateCollector128t` collector can be used for monitoring the admin, oper and redundancy status of various device-interfaces configured on the node. The name is available as `device-interface` tag and telegraf `tagpass` can be used to filter specific interfaces as needed. For example:
+```toml
```
agrawalkaushik, line 476: The new way of configuring this input should appear before the old way - that way any user referring to the doc for guidance with configuration is more likely to just use the new approach.

docs/plugin_monitoring_agent.md
```diff
@@ -488,11 +547,36 @@ The `peerPathStateCollector128t` collector can be used for monitoring the up/dow
...
+  ## Data format to consume.
+  ## Each data format has its own unique set of configuration options, read
+  ## more about them here:
+  ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md
+  data_format = "influx"
+
+  ## To filter on select device interfaces or network interfaces, you can use the `tagpass` and `tagdrop` concepts
+  ## from telegraf. For example:
+  ## [[inputs.exec.tagpass]]
+  ##     device-interface = ["wan1"]
```
agrawalkaushik, line 572: @BenMatase We should include all the different tags here for completeness.

docs/release_notes_monitoring_agent_1.X.md
```diff
@@ -1,7 +1,49 @@
 ---
-title: 128T Monitoring Agent 1.1 Release Notes
-sidebar_label: 1.1
+title: 128T Monitoring Agent 1.X Release Notes
+sidebar_label: 1.X
```
agrawalkaushik, line 3: The `1.2` should be in its own markdown file and add as a unique link in the sidebar

docs/plugin_monitoring_agent.md
gregschrock: This is the correct way to denote a new line in markdown. From the [Markdown: Syntax](https://daringfireball.net/projects/markdown/syntax#p) guide:

> When you do want to insert a `<br />` break tag using Markdown, you end a line with two or more spaces, then type return.

docs/plugin_monitoring_agent.md
gregschrock: Good point. I'll remove it.

docs/plugin_monitoring_agent.md
gregschrock: will update

sidebars.js
BenMatase: I can do that. I thought this would be nice since it would be less jumping around to see the different versions that are compatible with the 128T version. Consistency is important though

docs/plugin_monitoring_agent.md
BenMatase: This is the sample config so I'll open up an issue to include all of the tags in the sample  for next release

agrawalkaushik: Can we update the documentation for now and we'll update the default config.

docs/plugin_monitoring_agent.md
BenMatase: It looks like my vscode removed all of these. It looks the same to me?

docs/release_notes_monitoring_agent_1.X.md
BenMatase: Resolving because nothing to do. Thanks for the tips

docs/plugin_monitoring_agent.md
gregschrock: @BenMatase, I have a markdown linter running in my editor which actively removes newlines that have no effect. In this case, it (correctly according to markdown guidelines) puts `:::import ... :::import` all onto the same line which does break the Docusaurus behavior. So, you're correct that Docusaurus handles this case when there is no trailing whitespace. It's just a technicality we should follow to make everything function when the document is correctly linted.

More generally though, there were instances of the direct `<br/>` usage which should use this approach instead.

BenMatase: I talked to Greg and we agreed the current state is fine for now.

PR #70, opened 2020-06-09
Release notes for 128T 4.3.7

MichaelBaj:


docs/release_notes_128t_4.3.md
```diff
@@ -3,6 +3,27 @@ title: 128T 4.3 Release Notes
...
 
+## Release 4.3.7
+
+### Issues Fixed
+
+- **I95-24681** Grammatical improvements to HA initialization, providing more clarity around the use of specific IP addresses
+------
+- **I95-26276** Enabled OSPF authentication in configuration
+------
+- **I95-35172** DHCP Server configuration on network-interface is not real time configurable
```
ptimmons, line 14: This sounds worse than it presumably is. If the actual problem is that you cannot add DHCP configuration from scratch without a restart, then I suggest:
Adding DHCP server instances requires a software restart.

docs/release_notes_128t_4.3.md
```diff
@@ -3,6 +3,27 @@ title: 128T 4.3 Release Notes
...
+  Until the system is upgraded to 4.3.7, restarting the 128T will enable the DHCP server
+------
+- **I95-35401** SVR traffic would be dropped as a result of tenant members source type being incorrectly classified.
+  _**Conditions:**_ When the interface has an adjacency and Tenant members are applied via neighborhoods and/or child tenants. The tenant table will show the source type as `PUBLIC` for that entry when it should show as `HYBRID`
```
ptimmons, line 18: s/Tenant/tenant

This is really obtuse... would someone even really care to know this detail? Why not dumb it down to "Fixed a cosmetic issue where the tenant source table can incorrectly show PUBLIC instead of HYBRID under certain configuration scenarios."

docs/release_notes_128t_4.3.md
```diff
@@ -3,6 +3,27 @@ title: 128T 4.3 Release Notes
...
+------
+- **I95-35602** The command `show network-interface` may result in a `Unhandled TypeError` in the PCLI when a PPPoE interface is down
+------
+- **I95-35633** Configuration edits within the GUI have been dramatically increased
```
ptimmons, line 22: I presume this refers to performance and not quantity of edits :-)

Suggestion:
The GUI performance has been improved for configuration edit operations.

docs/release_notes_128t_4.3.md
MichaelBaj: This is not simply a cosmetic issue as SVR traffic is being dropped.  The "Conditions" explain the conditions for how the issue would manifest itself.

docs/release_notes_128t_4.3.md
MichaelBaj: Hah!  What a feature!  We make 10 edits for every one you create

PR #72, opened 2020-06-09
Updated documentation for monitoring agent 2.1.1 and 1.2.1

BenMatase:
* Updated references to 1.2.0 and 2.1.0 to 1.2.1 and 2.1.1
* Added mac address docs for device state collector and a version history

### DIFF
--- a/docs/plugin_monitoring_agent.md
+++ b/docs/plugin_monitoring_agent.md
@@ -7,13 +7,13 @@ The 128T Monitoring Agent is an entity for collecting data from a node running 1
 
 The monitoring agent at its core is designed to be able to push data to external platforms. It currently leverages the [Telegraf](https://www.influxdata.com/time-series-platform/telegraf/) collection stack on every 128T router. However, the monitoring agent is designed with other tools and scale in mind. The monitoring agent is composed of the following:
 
-- **monitoring-agent-cli**
+- **monitoring-agent-cli**  
   Used for configuring and interacting with the underlying application
 
-- **collectors**
+- **collectors**  
   A set of inputs designed to collect data from several sources such as metrics, events etc.
 
-- **targets**
+- **targets**  
   A set of outputs for collecting and pushing the data to various data sinks such as external monitoring platforms, files on disk etc.
 
 ## Installation
@@ -493,7 +493,7 @@ In versions 1.2.0, 2.1.0 and later, the more feature rich `t128_events` seen abo
 
 ### Device Interface State Collector
 
-The `deviceInterfaceStateCollector128t` collector can be used for monitoring the admin, oper and redundancy status of various device-interfaces configured on the node. The name is available as `device-interface` tag and telegraf `tagpass` can be used to filter specific interfaces as needed. For example:
+The `deviceInterfaceStateCollector128t` collector can be used for monitoring the admin, oper and redundancy status of various device-interfaces configured on the node. The device interface name is available as the `device-interface` tag and the mac address is available as the `mac-address` tag. Telegraf `tagpass` can be used to filter specific interfaces as needed. For example:
 
 ```toml
 [[inputs.exec]]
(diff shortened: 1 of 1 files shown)


BenMatase:
- [ Update dnf output once package is published ]

docs/plugin_monitoring_agent.md
```diff
@@ -28,51 +28,50 @@ The 128T Monitoring Agent can be obtained from the official 128T software reposi
 The agent can be install using dnf utility. For example.
 
 ```console
-# dnf install 128T-monitoring-agent
-Last metadata expiration check: 0:00:08 ago on Mon 18 May 2020 08:30:20 PM UTC.
+dnf install -y 128T-monitoring-agent
```
agrawalkaushik, line 31: nit: but probably not a good idea to include `-y` in a step like this - especially in documentation. Can we fix it to be without the `-y` - you can just copy paste some of the contents from the LHS instead of re-running the command etc.

docs/release_notes_monitoring_agent_1.2.md
```diff
@@ -3,44 +3,78 @@ title: 128T Monitoring Agent 1.2 Release Notes
...
+
+---
+
+- **MON-205** Honor the input enable/disable flag in the agent's config
+
+  _**Resolution**_ The configuration allows the user to disable an input. However, an input was being treated as enabled as long as it existed in the config.
+
+---
+
+- **MON-225** Correct bug where the `t128_events` input occasionally missed or delayed events
```
agrawalkaushik, line 32: Can we rename this to be:

`t128_events` input would occasionally drop or delay events

docs/release_notes_monitoring_agent_1.2.md
```diff
@@ -3,44 +3,78 @@ title: 128T Monitoring Agent 1.2 Release Notes
...
+- **MON-218** Expose MAC address in the device state input
+
+  Allow better correlation between device and network interfaces.
+
+---
+
+- **MON-210** Improve performance of several provided inputs
+
+  Reduce the resource consumption as well as the time needed to collect data. In particular, the `t128_metrics` input has been significanly improved.
+
```
agrawalkaushik, line 21: I think we should also include a blurb about the telegraf upgrade here as well - since it comes with new inputs etc. It would be helpful to know which version of telegraf we are on.

docs/release_notes_monitoring_agent_1.2.md
```diff
@@ -3,44 +3,78 @@ title: 128T Monitoring Agent 1.2 Release Notes
...
+- **MON-205** Honor the input enable/disable flag in the agent's config
+
+  _**Resolution**_ The configuration allows the user to disable an input. However, an input was being treated as enabled as long as it existed in the config.
+
+---
+
+- **MON-225** Correct bug where the `t128_events` input occasionally missed or delayed events
+
+  _**Resolution**_ Update the telegraf dependency for improvements and bug fixes. Particularly, the `execd` input had a bug that occasionally caused the `t128_events` issue.
```
agrawalkaushik, line 34: Can we just say:

Updated the telegraf dependency as well as update the `execd` input to better handle new lines

docs/release_notes_monitoring_agent_2.1.md
```diff
@@ -3,36 +3,73 @@ title: 128T Monitoring Agent 2.1 Release Notes
 sidebar_label: '2.1'
 ---
 
+## Release 2.1.1
+
+### New Features and Improvements
+
+---
+
```
agrawalkaushik, line 11: same comments as the 1.x version apply here as well.

docs/release_notes_monitoring_agent_1.2.md
gregschrock: The telegraf update does not include any new components - only bug fixes. I can note the version we're on. Our build also has a couple changes pulled from master, so it makes things a little less clear.

agrawalkaushik: I still think its important to highlight the upstream telegraf version in the release notes.

gregschrock: so you're saying they should be highlighted rather than just mentioned as I've done?

PR #73, opened 2020-06-15
Added documentation for the mosh plugin

laneshields:


docs/plugin_mosh.md
```diff
@@ -0,0 +1,18 @@
+# Mosh Plugin
```
MichaelBaj, line 1: Please add the `title` and `sidebar_label` front matter tags.

laneshields: Fixed

PR #86, opened 2020-07-02
Document Monitoring Agent 3.0 along with release notes

gregschrock:
## Description

Document the Monitoring Agent 3.0 release


docs/plugin_monitoring_agent.md
```diff
@@ -20,10 +20,11 @@ The monitoring agent at its core is designed to be able to push data to external
...
-| --------------------------- | --------------------------- |
-| 128T-monitoring-agent-1.2.1 | 128T >= 4.1.0; 128T < 4.3.0 |
-| 128T-monitoring-agent-2.1.1 | 128T >= 4.3.0               |
+| Monitoring Agent                | 128T                        |
+| ------------------------------- | --------------------------- |
+| **128T-monitoring-agent-3.0.0** | 128T >= 4.1.0               |
+| 128T-monitoring-agent-2.1.1     | 128T >= 4.3.0               |
+| 128T-monitoring-agent-1.2.1     | 128T >= 4.1.0; 128T < 4.3.0 |
 
 The agent can be install using dnf utility. For example.
```
MichaelBaj, line 29: Instead provide a link to the plugin installation guide as can be seen from the other plugin examples.

gregschrock: Unfortunately, although it's listed under `plugins`, the Monitoring Agent is not a plugin. That's in the works, but is not complete yet. Admins still need to directly `yum` or `dnf` install the agent.

MichaelBaj: Right you are.  I keep forgetting that.  Soon enough.

agrawalkaushik: The core monitoring agent is actually not a plugin and we dont have a plugin support around it yet. So the plugin install instructions wont apply here unfortunately. 

However, the 3.0 version is meant to be a single version that can work across all 128T software version, so maybe we should just remove all references to the 1.x and 2.x from the doc?

gregschrock: Great point. They shouldn't be installing other versions. I'll make that change.

docs/plugin_monitoring_agent.md
```diff
@@ -20,10 +20,11 @@ The monitoring agent at its core is designed to be able to push data to external
 
 The 128T Monitoring Agent can be obtained from the official 128T software repository. The following versions of the **monitoring-agent** are available for corresponding 128T software version.
 
-| Monitoring Agent            | 128T                        |
-| --------------------------- | --------------------------- |
-| 128T-monitoring-agent-1.2.1 | 128T >= 4.1.0; 128T < 4.3.0 |
-| 128T-monitoring-agent-2.1.1 | 128T >= 4.3.0               |
+| Monitoring Agent                | 128T                        |
+| ------------------------------- | --------------------------- |
+| **128T-monitoring-agent-3.0.0** | 128T >= 4.1.0               |
```
agrawalkaushik, line 25: @gregschrock I think its worth explicitly calling out in a foot note under the install section that the version is backward compatible to all 128T software version from 4.1.0 and greater and should be the preferred software version for new installation and upgrades.

docs/plugin_monitoring_agent.md
agrawalkaushik: @gregschrock so talked to @MichaelBaj offline about it a bit. I dont think we can get rid of the other references quite yet since they are still being used in the field. We can deprecate and remove references over time. For now we can just higlight the fact that 3.0 is the preferred version (see note above)

docs/plugin_monitoring_agent.md
agrawalkaushik: We should also either update the `dnf install` screenshot below for 3.0 or just remove it altogether (probably the latter).

docs/plugin_monitoring_agent.md
```diff
@@ -197,23 +204,23 @@ Path: `/var/lib/128t-monitoring/inputs/t128_metrics.conf`
...
-    ##     latency = "stats/bfd/peer-path/latency"
+    ##     packets-received = "stats/aggregate-session/service/packets-received"
     ##
-    ## [inputs.t128_metric.metric.parameters]
+    ## [inputs.t128_metrics.metric.parameters]
     ##     parameter_name = ["value1", "value2"]
-    ##     peer_path = ["path1"]
+    ##     service = ["service1"]
 
     timeout = "15s"
```
agrawalkaushik, line 219: Not that it hurts to keep it at 15s but given all the new metric collector improvements, this might not be necessary anymore.

docs/plugin_monitoring_agent.md
```diff
@@ -305,10 +312,47 @@ In this example data is sent via syslog:
...
+
+Within an **input** configuration, several variables have been made available for substitution.
+
+| Value             | Meaning                                      | Version Introduced |
+| ----------------- | -------------------------------------------- | ------------------ |
+| `${ROUTER}`       | The router name of the running 128T instance | 3.0.0              |
+| `${NODE}`         | The node name of the running 128T instance   | 3.0.0              |
+| `${128T_VERSION}` | The version of the running 128T instance     | 3.0.0              |
+
+An example of this would be (note that the `example` input does not exist).
```
agrawalkaushik, line 329: nit: the "note that the `example` input does not exist" is a bit distracting when reading this. Can we just remove it since its clear that its an example.

docs/plugin_monitoring_agent.md
```diff
@@ -441,6 +485,32 @@ The `metricCollector128t` python executable is responsible for collecting the co
 
 The `name` becomes the name of the measurement in the context of influxdb format. The `metric.fields` represent the various metrics to be collected. The `packets-received` in the above example will be field-name for the `stats/aggregate-session/service/packets-received` KPI which is the path of that KPI from the 128T REST API documentation. Finally, the `metric.parameters` can be used to configure key parameters such as `service` to be used for filtering the set of collected stats. In the above example, the metrics would be collected for all services but the `service` parameter can be used to specify a subset of services to monitor instead.
```
agrawalkaushik, line 486: For the "influxdb format" part we should include some external link to telegraf/influx documentation explaining what that format is. Just as a way to educate the reader if they are curious.

agrawalkaushik: For the "128T REST API documentation" we should add a link to (api_rest_4.2.0.md)

docs/plugin_monitoring_agent.md
```diff
@@ -441,6 +485,32 @@ The `metricCollector128t` python executable is responsible for collecting the co
...
 
+A custom set of metrics can be collected by configuring the `t128_metrics` input as described in the sample. The configuration follows the same structure as the default file, but the metrics are nested under the input.
+
+```console
```
agrawalkaushik, line 490: We already captured the default config example in a previous section. In fact the above statement talks about a custom set of metrics so its probably better to include an example of how this would look like with a set of custom metrics in place.

docs/plugin_monitoring_agent.md
gregschrock: Good call. I'll be explicit and remove the install section. I don't think the install view is particularly beneficial, though I may leave the command itself without the console output.

docs/plugin_monitoring_agent.md
gregschrock: That's true. I don't think it's necessary. But it also shouldn't hurt anything and doesn't preclude a smaller interval.

docs/plugin_monitoring_agent.md
```diff
@@ -26,56 +26,14 @@ The 128T Monitoring Agent can be obtained from the official 128T software reposi
 | 128T-monitoring-agent-2.1.1     | 128T >= 4.3.0               |
 | 128T-monitoring-agent-1.2.1     | 128T >= 4.1.0; 128T < 4.3.0 |
 
-The agent can be install using dnf utility. For example.
+:::important  
+Monitoring Agent 3.X deprecates prior releases and is compatible with all previously supported 128T versions. It should be preferred for new installations and upgrades.  
+:::important
```
MichaelBaj, line 31: I'm not sure what happens when you close an admonition with another heading.  Proper form is:

:::admonition
text
:::

gregschrock: Good catch. I'm running the development server and it does render correctly. I'll remove.

- [ ] Remove closing admonition

agrawalkaushik:
Wait till we are ready to publish the RPMs

docs/plugin_monitoring_agent.md
```diff
@@ -660,3 +658,257 @@ The **top sessions** input above can be used to stream the view of the top 10 se
...
+
+- **MON-230** Make 3.X version of the Monitoring agent compatible with 4.1.0 <= 128T < 6.0.0
+- **MON-233** Upgrade telegraf to 1.14.5
+- **MON-234** Improve metrics collection performance by creating a native Telegraf plugin
+- **MON-198** Provide sample and push interval overrides per input in the agent's config
+- **MON-246** Enable value substitution in telegraf configuration files
+
+#### Issues Fixed
+
+- **MON-280** Make the arp state collector compatible with 128T 5.X
```
BenMatase, line 676: This is the only one that isn't in the format that the others are following where you have a problem and a resolution. Something like?
```
- **MON-280** Arp state collector does not produce metrics with 128T 5.X.

  _**Resolution**_ The arp state collector now dynamically handles data collection depending on the 128T version.
```
It isn't much more information, but just to be consistent.

BenMatase: No need to block for this

agrawalkaushik: done

PR #87, opened 2020-07-06
Move the ipsec-client plugin release notes to the primary doc

agrawalkaushik:
DOCS-115 #time 30m


docs/plugin_ipsec_client.md
```diff
@@ -58,7 +49,7 @@ exit
 ```
 
 :::note 
-This plugin can only connect to IPSEC endpoints that support pre-shared key authentication.
```
BenMatase, line 61: ty

PR #90, opened 2020-07-08
Release notes for 128T 4.4.1

MichaelBaj:


docs/release_notes_128t_4.4.md
```diff
@@ -3,6 +3,58 @@ title: 128T 4.4 Release Notes
 sidebar_label: 4.4
 ---
 
+## Release 4.4.1
+
+:::info
```
ptimmons, line 8: Upgrading to the 4.4.1 release requires version 2.6.0 or newer of the 128T installer. If you are upgrading a managed router using the conductor, the installer will be upgraded automatically. [[Fact check this]] If you are upgrading a router or conductor manually, please ensure you meet the minimum installer requirements before starting.

docs/release_notes_128t_4.4.md
```diff
@@ -3,6 +3,58 @@ title: 128T 4.4 Release Notes
...
+The minimum 128T-installer version of 2.6.0 is required for the 4.4.1 update. (rpm package: 128T-installer-2.6.0-1.x86_64.rpm)
+:::
+
+### Issues Fixed
+
+- **I95-33403** REST APIs have been added for routing show commands
```
ptimmons, line 14: ... for retrieving information about the 128T routing protocols.

docs/release_notes_128t_4.4.md
```diff
@@ -3,6 +3,58 @@ title: 128T 4.4 Release Notes
...
+------
+- **I95-34649** `best-effort` path handling for `proportional` load balancing is not honored by service-policy
+------
+- **I95-34751** LTE certified to run on Verizon wireless networks
+------
+- **I95-35038** Configuration Explorer within GUI can sometimes produce a "Something went wrong" message when searching or scrolling
+------
+- **I95-35313** Startup delay of 128T when many peer paths exist
+------
+- **I95-35354** Downloading a software image during or immediately prior to an upgrade will incorrectly complete the upgrade if the download was successful before the upgrade starts
```
ptimmons, line 26: You had me until the "if the download was successful before the upgrade starts" bit. This needs clarification.

docs/release_notes_128t_4.4.md
MichaelBaj: This is a correct statement.  Fact checked.  Simply upgrading through the GUI will ensure that the installer is upgraded automatically.

docs/release_notes_128t_4.4.md
MichaelBaj: There exists an unlikely race condition wherein the successful return code of a download operation (that happens asynchronously) causes an upgrade in progress to terminate prematurely.

Proposed rewording:
"Downloading a software image prior to, or during an upgrade, can terminate the upgrade operation if the download completed successfully prior to the completion of the upgrade"

PR #140, opened 2020-10-07
updated release notes for GRE 1.1.3 and 2.1.3

santi969:


docs/plugin_gre.md
```diff
@@ -552,3 +552,14 @@ In addition, a `ping-monitor` service is started for each configured tunnel, the
 - **PLUGIN-677** GRE plugin doesn't start up correctly post reboot
 
     _**Resolution**_ Implemented a config watcher service to handle startup conditions and dynamically apply configuration changes at runtime.
+
+### Release 1.1.3, 2.1.3
+
+#### Issues Fixed
+- **PLUGIN-799** Pillar data for for GRE tunnel is false by default
+
+    _**Resolution**_ Set denabled default to True when generating gre pillar data
```
MichaelBaj, line 561: Typo: "enabled"

MichaelBaj (changes requested):

docs/plugin_gre.md
Chr1st0ph3rTurn3r, line 561: Also wondering if it should be "Set default to True..." rather than "Set enabled default to True..."

docs/plugin_gre.md
```diff
@@ -552,3 +552,14 @@ In addition, a `ping-monitor` service is started for each configured tunnel, the
...
+
+- **PLUGIN-479** Address range checking not valid for GRE tunnel
+
+    _**Resolution**_ Use non-strinct mode when getting plugin network in config generation
```
Chr1st0ph3rTurn3r, line 565: is "strinct" a typo? should it be "non-strict" mode?

docs/plugin_gre.md
santi969: How's "Set default to True (for "enabled" key) when generating gre pillar data" sound? Or you think I should get rid of enabled completely

Chr1st0ph3rTurn3r: I think that sounds great - makes sense that way.

MichaelBaj (approved):

docs/plugin_gre.md
```diff
@@ -552,3 +552,14 @@ In addition, a `ping-monitor` service is started for each configured tunnel, the
 - **PLUGIN-677** GRE plugin doesn't start up correctly post reboot
 
     _**Resolution**_ Implemented a config watcher service to handle startup conditions and dynamically apply configuration changes at runtime.
+
+### Release 1.1.3, 2.1.3
+
+#### Issues Fixed
+- **PLUGIN-799** Pillar data for GRE tunnel is false by default
+
+    _**Resolution**_ Set default to True (for "enabled" key) when generating gre pillar data
```
agrawalkaushik, line 561: This sounds a little contrived, can we say this instead

Set the enabled key to true by default when generating the pillar data

Dropped "key" as thats an implementation detail and "GRE" since the context makes that obvious.

agrawalkaushik (changes requested):

MichaelBaj (approved):

agrawalkaushik (approved):

jeffmart (approved):

PR #151, opened 2020-11-23
Use screen for upgrade and rollback

migolnikov:
Add instructions to use Screen during interactive installer upgrade and rollback


migolnikov:
This change is also needed in the Installer Command Line Reference. I will try to do that next

Chr1st0ph3rTurn3r (approved):

MichaelBaj (approved):

PR #159, opened 2021-01-04
Added services description document

ptimmons:
As requested by Andy Epstein and Brad Jones (JNPR legal).

### DIFF
--- a/docs/about_services_description.md
+++ b/docs/about_services_description.md
@@ -1,0 +1,180 @@
+---
+title: Services Description
+sidebar_label: Services Description
+description: 'This document describes the maintenance and support services available to 128 Technology's end customers.'
+---
+
+## Introduction and Eligibility
+
+This Services Description Document (“SDD”) describes the maintenance and support services (“Services”) available to licensed subscribers (“Customers”) of 128 Technology software products (“Software”). 
+
+The Services are subject to the terms of this SDD and the 128 Technology End User License Agreement or other written agreement covering within its scope the terms and conditions under which Customer has purchased a license subscription to the Software (“Subscription Contract”).
+
+The Services will be rendered by 128 Technology (or its representatives or contractors) to the Customer as identified by name and address in the order for the license subscription to the Software (whether purchased directly from 128 Technology or by a reseller or distributor authorized by 128 Technology to resell subscription licenses for Software). In order to be eligible to receive Services, the Customer must hold a valid Subscription Contract.
+
+## Service Features and Deliverable Description
+
+As part of the Services, 128 Technology will use commercially reasonable efforts to provide Customer with the following during the paid subscription period stated in the applicable Subscription Contract (“Service Period”):
+
+### Technical Support
+
+Customer will have access to 128 Technology technical support engineers by phone and online 24/7/365 and without restriction as to the number of cases Customer may open. As a single point of contact for all support needs, 128 Technology technical support engineers will help diagnose and troubleshoot products and provide workaround solutions where necessary. 128 Technology technical support engineers will provide assistance with product use and Software Patches, Software Updates, and Software Upgrades, each as defined below. Services availability for any particular release of Software is subject to limitations and restrictions under [128 Technology’s Product Software Lifecycle Support Policy](about_support_policy.md).
+
+For critical or service-impacting problems, Customers are requested to contact 128 Technology technical support by calling **1-781-203-8400**. For non-critical/non-service-impacting problems, Customers can open cases through 128 Technology’s online support portal.
+
+For additional details about support procedures and case priority guidelines, please refer to Schedule 1 (ERROR PRIORITY LEVELS/RESPONSE GUIDELINES) attached hereto.
+
+### Software Releases
+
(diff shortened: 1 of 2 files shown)


docs/about_services_description.md
```diff
@@ -0,0 +1,180 @@
...
+title: Services Description
+sidebar_label: Services Description
+description: 'This document describes the maintenance and support services available to 128 Technology's end customers.'
+---
+
+## Introduction and Eligibility
+
+This Services Description Document (“SDD”) describes the maintenance and support services (“Services”) available to licensed subscribers (“Customers”) of 128 Technology software products (“Software”). 
+
+The Services are subject to the terms of this SDD and the 128 Technology End User License Agreement or other written agreement covering within its scope the terms and conditions under which Customer has purchased a license subscription to the Software (“Subscription Contract”).
```
Chr1st0ph3rTurn3r, line 11: Globally, Customer should be prefaced with an article - the Customer, a Customer. Or, it should be changed to Customers, as indicated in the first paragraph. 
There are several instances where this must be looked at individually, based on sentence structure.

docs/about_services_description.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Software Releases
+
+During the Service Period, 128 Technology shall make available to Customer the Updates (defined below) of the Software covered under a Subscription Contract, but solely for Customer’s use as a replacement of a different release of the Software duly licensed to such Customer. Any such Update shall be subject to the terms and conditions set forth below:
+
+I. Rights in Updates. An Update downloaded by or otherwise delivered to Customer under a Subscription Contract may be used only to replace Software licensed to Customer and only if Customer agrees not to use the Update until it first discontinues use of the replaced Software and makes it unavailable for any use, other than for archival use. Once an Update is installed and placed in use and all copies of the replaced release of the Software is removed from all use, the Customer’s rights in the Update will be subject to the same terms, conditions, and restrictions as the Software originally licensed to Customer, as set forth in the applicable Subscription Contract.
```
Chr1st0ph3rTurn3r, line 31: "Software licensed to Customer and only if Customer agrees not..." 
In this case customer should be prefaced with an article - "licensed to the Customer, and only if the Customer agrees..."

sidebars.js
```diff
@@ -7,6 +7,7 @@ module.exports = {
       "about_supported_platforms",
       "about_certified_platforms",
       "about_security_policy",
+      "about_services_description"
```
MichaelBaj, line 10: missing comma at the end of this list, causing the "build" to fail

docs/about_services_description.md
MichaelBaj, line 31: I'm presuming this statement came from legal, so I don't imagine we should change it. Maybe I'm being too technical, but older versions of the software will exist for the purposes of rolling back.

docs/about_services_description.md
```diff
@@ -0,0 +1,180 @@
...
+
+Customer may from time to time submit requests or suggestions for feature or functionality enhancements that would require future development. 128 Technology may review and discuss such requests with Customer on a periodic basis, as mutually determined by the parties. 128 Technology shall have no obligation to develop or deliver any enhancements.
+
+**Escalation**
+
+Customer and 128 Technology may establish and periodically update an escalation list of management employees responsible for the support of the Software. Use of such list will be subject to compliance with 128 Technology’s guidelines and policies, including those applicable to privacy.
+
+**Service Level Targets**
+
+The response times and other service levels described in this Schedule 1 are targets or objectives that 128 Technology strives to achieve consistently in providing Services to Customers. A failure to meet any one or more of the service level targets will not create any liability on the part of 128 Technology, nor a right to any credits or to withhold payment on the part of Customers.
```
Chr1st0ph3rTurn3r, line 180: "described in this Schedule 1 are..." 
This is a bit funky. "described in Schedule 1 are..." or "described in this Schedule are..."?

Chr1st0ph3rTurn3r (changes requested):

MichaelBaj (changes requested):

docs/about_services_description.md
ptimmons: This document was copied and pasted verbatim from Juniper corporate legal and Andy Epstein. Unless this is a hard requirement of yours I suggest we leave it as-is.

docs/about_services_description.md
ptimmons: This document was copied and pasted verbatim from Juniper corporate legal and Andy Epstein. Unless this is a hard requirement of yours I suggest we leave it as-is.

docs/about_services_description.md
ptimmons: This document was copied and pasted verbatim from Juniper corporate legal and Andy Epstein. Unless this is a hard requirement of yours I suggest we leave it as-is.

docs/about_services_description.md
ptimmons: Correct, it came from legal. I suggest we leave it as-is.

sidebars.js
ptimmons: Fixed.

MichaelBaj (approved):

MichaelBaj:
The build failed due to

```
Creating an optimized production build...
Error: Error while parsing markdown front matter.
This can happen if you use special characteres like : in frontmatter values (try using "" around that value)
can not read a block mapping entry; a multiline key may not be an implicit key at line 5, column 1:
```

I suspect this is because you used a single quote instead of a double quote for the description.

PR #185, opened 2021-02-05
Updated 4.5 release notes with 4.5.5, I95-38833.

Chr1st0ph3rTurn3r:


docs/release_notes_128t_4.5.md
```diff
@@ -3,6 +3,12 @@ title: 128T 4.5 Release Notes
 sidebar_label: 4.5
 ---
 
+## Release 4.5.5
+
+### Resolved Issues
+
+- **I95-38833 BGP is not coming up on a WAN interface with Q-in-Q enabled:** Resolved an issue where a WAN interface with Q-in-Q enabled was incorrectly reading the VLAN tag from the outer VLAN header.
```
MichaelBaj, line 10: BGP was just a coincidence of the failure mode.  The actual issue has to do with DSCP marking affecting our parser.  The BGP packets were marked with a non-zero DSCP value which meant that we could not properly identify the Q-in-Q packets. Suggested rewording:
I95-38833 Q-in-Q packets are not recognized when packets have non-zero DSCP value.
- you can keep the additional text as is.

MichaelBaj (changes requested):

docs/release_notes_128t_4.5.md
Chr1st0ph3rTurn3r: Updated.

MichaelBaj (approved):

jeffmart (approved):

PR #193, opened 2021-03-01
Document Monitoring Agent 3.3.1

gregschrock:
## Description

Document the changes to the Monitoring Agent in 3.3.1. This is scheduled to release with 128T 5.1.0.


docs/plugin_monitoring_agent.md
```diff
@@ -20,9 +20,9 @@ The monitoring agent at its core is designed to be able to push data to external
 
 The 128T Monitoring Agent can be obtained from the official 128T software repository. The following versions of the **monitoring-agent** are available for corresponding 128T software version.
 
-| Monitoring Agent                | 128T                        |
-| ------------------------------- | --------------------------- |
-| **128T-monitoring-agent-3.0.0** | 128T >= 4.1.0               |
+| Monitoring Agent                | 128T          |
+| ------------------------------- | ------------- |
+| **128T-monitoring-agent-3.X.X** | 128T >= 4.1.0 |
```
agrawalkaushik, line 25: Why `X.X` - that doesn't make sense. I think this table was created when we had multiple versions. We should just get rid of this section instead of saying `3.X.X`

docs/plugin_monitoring_agent.md
```diff
@@ -40,12 +40,14 @@ dnf install 128T-monitoring-agent
 
 | Release      | Modification                                                    |
 | ------------ | --------------------------------------------------------------- |
+| 3.3.1        | `name` differentiator                                           |
```
agrawalkaushik, line 43: `differentiator` sounds odd - Can we call it runtime environment name instead?

docs/plugin_monitoring_agent.md
```diff
@@ -68,6 +70,8 @@ outputs:
   - name: message_queue
 ```
 
+When using multiple instances of the Monitoring Agent, specify the `name` to uniquely identify each instance. Without this distinction, two inputs with the same name can colide (be started/stopped/reconfigured by both instances) even if they are listed in separate `config.yaml` files. This field may be ignored when using the default `config.yaml` file. In other cases, the name of the directory containing the config file will be used as the name if none is specified.
```
agrawalkaushik, line 73: `colide` -> `collide`

agrawalkaushik: I'd recommend a bit of re-ordering of this paragraph. Something like this instead

```
When multiple instances of monitoring agent are running...(basically define the problem). By default, the name of the directory containing the config fill will be used as the name of the runtime environment. Alternatively, the `name` field in the config can be used to identify a unique runtime environment as well.
```

docs/plugin_monitoring_agent.md
```diff
@@ -276,11 +280,12 @@ For syslog output, not specifying the `default_sdid` parameter can result in emp
...
-| ----------------- | -------------------------------------------- | ------------------ |
-| `${ROUTER}`       | The router name of the running 128T instance | 3.0.0              |
-| `${NODE}`         | The node name of the running 128T instance   | 3.0.0              |
-| `${128T_VERSION}` | The version of the running 128T instance     | 3.0.0              |
+| Value             | Meaning                                           | Version Introduced |
+| ----------------- | ------------------------------------------------- | ------------------ |
+| `${ROUTER}`       | The router name of the running 128T instance      | 3.0.0              |
+| `${NODE}`         | The node name of the running 128T instance        | 3.0.0              |
+| `${128T_VERSION}` | The version of the running 128T instance          | 3.0.0              |
+| `${WEB_PORT}`     | A local port that can be used to access 128T APIs | 3.0.0              |
```
agrawalkaushik, line 288: is `3.0.0` correct and we just forgot to update the doc at that time?

docs/plugin_monitoring_agent.md
```diff
@@ -659,8 +703,83 @@ The **top sources** input can be used to capture the list of source IP addresses
...
+- **MON-311** A `t128_graphql` input is now available
+- **MON-311** A dedicated `t128_device_state` input is now available
+- **MON-311** A dedicated `t128_arp_state` input is now available
+- **I95-38959** Improve `t128_metrics` performance with bulk retrieval
+
+#### Issues Fixed:
+
+- **MON-316** Correct default metrics bug where interface `received-missed` was collected from `stats/interface/received/error`
+
+  _**Resolution**_ Point `received-missed` to the correct `stats/interface/received/missed`
```
agrawalkaushik, line 781: This resolution is a bit misldeading. The resolution was that the default metrics config was updated to point...

We did not do this generically for all configuration - so if a user copy/pasted that config to create their own, its likely still wrong.

docs/plugin_monitoring_agent.md
gregschrock: - [ ] Remove version compatibility table

docs/plugin_monitoring_agent.md
gregschrock: - [ ] rename `differentiator`

docs/plugin_monitoring_agent.md
gregschrock: - [ ] colide typo
- [ ] restructure description of the environment `name`

docs/plugin_monitoring_agent.md
gregschrock: It has been there since we released 3.0, but I think we deliberately chose not to publicize it at that point.

docs/plugin_monitoring_agent.md
gregschrock: good point 😆 

- [ ] reword default metrics resolution

BenMatase (approved):

agrawalkaushik (approved):

docs/plugin_monitoring_agent.md
```diff
@@ -68,6 +66,8 @@ outputs:
   - name: message_queue
 ```
 
+When multiple instances of the Monitoring Agent are running, two inputs with the same name can collide (be started/stopped/reconfigured by both instances) even if they are listed in separate `config.yaml` files. By default, the name of the directory containing the config file will be used as the name of the environment in order to distinguish them. Alternatively, the config `name` field can be explicitly specified to identify the unique environment.
```
MichaelBaj, line 69: Is this for development purposes only? A customer should not be able to launch multiple instances of the monitoring-agent, correct?

MichaelBaj (approved):

docs/plugin_monitoring_agent.md
agrawalkaushik: Mike, this was done mainly to allow using the monitoring agent for development. However, its implemented in a way where customers could use this to create separate environments if they chose for some reason. The end effect is the same since we use the environment's to launch telegraf instances. For the most part I would imagine customers needing just the default environment.

salegria8 (approved):

PR #197, opened 2021-03-02
OTP ISO install

Chr1st0ph3rTurn3r:


docs/intro_installation_bootable_media.md
```diff
@@ -22,21 +23,28 @@ After imaging the ISO onto removable media, insert it into the target machine an
...
+
+![VGA Boot with Interactive Install](/img/install_select_interactive.png)
 
-![Boot Screen](/img/intro_installation_bootable_media_boot.png)
+Differences for the Serial console are described in the Appendix. 
+
+To perform the installation using the OTP Installation, use the arrow keys to select either `OTP Install 128T Routing Software Serial Console` or `OTP Install 128T Routing Software VGA Console`. The OTP Installation for the serial or VGA console is described in the section [Router Installation Using OTP](intro_otp_iso_install.mdx).
 
 :::note
 Not all hardware has video support, therefore booting to console is the default (a console user may not be able to select an option). The default option is selected after a 30 second timeout.
```
jeffmart, line 37: I would like to have this section updated. The serial console access should normally work. The Serial console default size expected is 80x25. If the terminal being used to connect to the console is less than this row(80) X column(25), the display may appear to the console user that they are unable to select the entries or the display does not repaint correctly. Also in the initial bootup, of the console is disconnected, there is no repaint capability, the arrow up and down buttons should be used to show the available selections. Maybe re-word something like:

"Not all hardware has video support, therefore booting to the serial console is the default and will be automatically selected after 30 seconds. When using the serial console, the terminal size is 80x25 (anything smaller may result in abnormal navigation behavior."

docs/intro_installation_bootable_media.md
```diff
@@ -22,21 +23,28 @@ After imaging the ISO onto removable media, insert it into the target machine an
...
+
+Selecting the wrong type of console (Serial or VGA) may result in garbage characters being displayed and the install hanging. If this is the case, reboot the target system and select the correct line for the target hardware.
 ::: 
+However, the procedure that follows here is the Interactive install on the VGA Console.
```
jeffmart, line 41: Remove "However,"

docs/intro_installation_bootable_media.md
```diff
@@ -22,21 +23,28 @@ After imaging the ISO onto removable media, insert it into the target machine an
...
 
-![Boot Screen](/img/intro_installation_bootable_media_boot.png)
+Differences for the Serial console are described in the Appendix. 
+
+To perform the installation using the OTP Installation, use the arrow keys to select either `OTP Install 128T Routing Software Serial Console` or `OTP Install 128T Routing Software VGA Console`. The OTP Installation for the serial or VGA console is described in the section [Router Installation Using OTP](intro_otp_iso_install.mdx).
 
 :::note
 Not all hardware has video support, therefore booting to console is the default (a console user may not be able to select an option). The default option is selected after a 30 second timeout.
+
+Selecting the wrong type of console (Serial or VGA) may result in garbage characters being displayed and the install hanging. If this is the case, reboot the target system and select the correct line for the target hardware.
```
jeffmart, line 39: The system actually does not hang, and if left will continue to load the OS to it's compleation. However the kernel default console will be set wrong so the the above resolution is still the recommended solution.

docs/intro_installation_bootable_media.md
jeffmart: I did not see the section below calling out the serial console until after I wrote the above, do we move the NOTE to the section below or add the last sentence above to below? (either works for me).

docs/intro_installation_bootable_media.md
```diff
@@ -141,6 +151,31 @@ After installing the 128T Software it is important to verify that the installati
...
+
+![Boot Screen](/img/intro_install_LegacyInstall.png)
+
+The following is the  installation screen when booting to a UEFI console:
+
+![New Boot Screen](/img/intro_install_OTPInstall_1.png)
+
+### Serial Console Troubleshooting
+
+In order to avoid SSH session timeout during installation, it is strongly recommended to use the **Screen** utility when performing an installation. The serial screen size must be a minimum of 80x25, otherwise the output is not readable.
```
jeffmart, line 166: I think we should change this section. The serial console access and behavior is really dependant on the host OS and utilities. For example I have never seen issues when using Windows with putty on screen display corruption (so using screen and ssh in this case does not apply). However I have seen different issues based on the different hardware using the serial console with OSX/screen, and these are different from hardware platform to hardware platform being imaged. With that said, I think we can reword this to be something like this:

Reiterate the terminal window size of 80x25, "if this is not the minumum size the console navigation may have unintended consequences or may not appear to navigate at all."

"When performing an installation via the serial console, some systems do not interpret control characters that may be passed on the serial console line. For example, the following may be seen during the installation process: [keep the png that you have there]"

docs/intro_installation_bootable_media.md
```diff
@@ -141,6 +151,31 @@ After installing the 128T Software it is important to verify that the installati
...
+In order to avoid SSH session timeout during installation, it is strongly recommended to use the **Screen** utility when performing an installation. The serial screen size must be a minimum of 80x25, otherwise the output is not readable. 
+
+![Corrupt Serial Install Complete](/img/install_serial_corrupt.png)
+
+In cases where the **Screeen** output becomes unreadable, use the following procedure. 
+1. Type: `^a` (Ctrl-a)
+2. Type: `k`
+3. Type: `y` [for yes to exist screen]
+4. Reconnect with screen (i.e., Screen command could be “screen /dev/tty.usbserial-1430 115200” ).
+5. Type: `^l` (Ctrl-l) (lower case L) to repaint the screen.
```
jeffmart, line 175: Add a sub note here stating that the repaint operation does not work on the initial boot screen to image a system so if the serial console is disconnected, it is recommended to restart the system and imaging process again.

docs/intro_installation_bootable_media.md
```diff
@@ -48,42 +56,44 @@ After the Linux installation completes, the 128T software installation begins. N
...
-   | Username | Password   |
-   | -------- | ---------- |
-   | root     | 128tRoutes |
-   | t128     | 128tRoutes |
+   | Username | Password   | Access Scope |
+   | -------- | ---------- | ------------ |
+   | root     | 128tRoutes | Linux account |
+   | t128     | 128tRoutes | Linux account |
+   | admin | 128Tadmin | 128T PCLI admin account |
+   | user | 128Tuser | 128T user PCLI account |
```
jeffmart, line 70: nit change "128T user PCLI account" to "128T PCLI user account" I may have provided the initial wrong content :)

jeffmart: actually, thinking about this more, it is PCLI and UI so remove the PCLI completely for both and use "128T admin account" and "128T user account"

docs/intro_installation_bootable_media.md
```diff
@@ -48,42 +56,44 @@ After the Linux installation completes, the 128T software installation begins. N
...
+The 128T Initializer tunes your operating system, prepares the platform to run the 128T software, and creates the bootstrapping files necessary to load the software. The Initializer is launched after the installation reboot. 
 
 1. On the 128T Initializer wizard screen, use the space bar to select either a **Router** or **Conductor **role for the 128T node and press the **Enter** key to select **OK**.
-  ![128T Role](/img/intro_install_initializer_role.png)
+
+  ![128T Role](/img/initializer_Serial2.png)
 
 2. For 128T routers, you will be prompted for the IP address(es) of your 128T conductor(s). If you have conductors, Enter their administrative addresses here, and this node will retrieve its configuration from the conductor. If you have only one conductor (i.e., a standalone conductor), leave the field labeled 2nd Conductor Address blank. If you have no conductors, choose **Skip**.
 3. When asked _What kind of Router node is this?_, select from the following options:
```
jeffmart, line 83: We might want to update this to reference the screen disaply. The text says router the screen shot has Conductor, maybe do "_What kind of Conductor/Route node is this?_

docs/intro_installation_bootable_media.md
```diff
@@ -48,42 +56,44 @@ After the Linux installation completes, the 128T software installation begins. N
...
 
-  ![Identify the Node](/img/intro_install_initializer_HASetup.png)
+  ![Identify the Node](/img/initializer_Serial3.png)
 
 - **Standalone:** This router has no highly available peer, and is not currently planned for high availability.
 - **1st HA Node:** This router is the first node of a high availability pair. You will be prompted for the High Availability address of this node. Note: The address _must_ be reachable by the 2nd HA Node.
```
jeffmart, line 88: There has been some confustion on this in the past. While we are in here, please update with the following:
"**1st HA Node:** This router is the first node of a high availability pair. You will be prompted to provide the local IP address for this node. The 2nd HA node will contact this node at the address provided to synchronize state. Note: The 1st Node IP address _must_ be reachable by the 2nd HA Node."

docs/intro_installation_bootable_media.md
```diff
@@ -48,42 +56,44 @@ After the Linux installation completes, the 128T software installation begins. N
...
+
+  ![HA Address](/img/initializer_Serial4.png)
+
 - **2nd HA Node:** This router is the second node of a high availability pair, where the first node has been initialized. You will be prompted for the High Availability address of the first node.
```
jeffmart, line 92: Update this with:
"**2nd HA Node:** This router is the second node of a high availability pair, where the first node has been initialized. You will be prompted to provide the 1st Node IP address for this 2nd node that will be used to synchronize state. Note: The 2nd Node IP address _must_ be reachable by the 1st HA Node."

docs/intro_installation_bootable_media.md
```diff
@@ -141,6 +151,31 @@ After installing the 128T Software it is important to verify that the installati
 
 4. Close the command prompt window. 
 
+## Serial Console Installation Information
+
+Please note that Serial and UEFI consoles display differently. The following is the  installation screen when booting to a Serial console:
```
jeffmart, line 156: Change "Serial and UEFI" to "Legacy and UEFI". Serial is not directly related as the difference are also similar in VGA installations for Legacy and UEFI. It will be obvious when someone performs the installation on different systems.

jeffmart (changes requested):
Please see individual comments below

docs/intro_installation_bootable_media.md
```diff
@@ -48,42 +56,44 @@ After the Linux installation completes, the 128T software installation begins. N
...
+  ![HA Address](/img/initializer_Serial4.png)
+
 - **2nd HA Node:** This router is the second node of a high availability pair, where the first node has been initialized. You will be prompted for the High Availability address of the first node.
 
 4. Enter the following system properties on the Node Info screen:
 
-   ![Node Information](/img/intro_install_initializer_nodeinfo.png)
+   ![Node Information](/img/initializer_Serial5.png)
 
 - **Node Name:** The unique name of the system within your 128T Router or Conductor, for example _labsystem1_. By default this field uses the Linux system's hostname.
```
jeffmart, line 98: While this is not specifically part of this changed we are changing the screen shots so this gives us the opertunity to also provide cascading examples. With the screen shots I provided out of band we should change the context of the nodeName.router/conductorName to match the screen shots which would be:

conductor-node1.conductor

so labsystem1 which be changed to "conductor-node1" and the router/conductorName would be changed to "conductor" to match the screen shot.

docs/intro_installation_bootable_media.md
Chr1st0ph3rTurn3r: done

docs/intro_installation_bootable_media.md
Chr1st0ph3rTurn3r: done

docs/intro_installation_bootable_media.md
Chr1st0ph3rTurn3r: done

docs/intro_installation_bootable_media.md
Chr1st0ph3rTurn3r: done

docs/intro_installation_bootable_media.md
Chr1st0ph3rTurn3r: done

docs/intro_installation_bootable_media.md
Chr1st0ph3rTurn3r: done

docs/intro_installation_bootable_media.md
Chr1st0ph3rTurn3r: done

docs/intro_installation_bootable_media.md
Chr1st0ph3rTurn3r: done

docs/intro_installation_bootable_media.md
Chr1st0ph3rTurn3r: done

docs/intro_installation_bootable_media.md
Chr1st0ph3rTurn3r: done

jeffmart (dismissed):

docs/intro_installation_bootable_media.md
```diff
@@ -141,6 +169,37 @@ After installing the 128T Software it is important to verify that the installati
...
+
+![Boot Screen](/img/intro_install_LegacyInstall.png)
+
+The following is the  installation screen when booting to a UEFI console:
+
+![New Boot Screen](/img/intro_install_OTPInstall_1.png)
+
+### Serial Console Troubleshooting
+
+- The serial screen size **must** be a minimum of 80x25, otherwise the output may not be readable, and the console navigation may have unintended consequences or may not appear to navigate at all.
```
MichaelBaj, line 184: The disclaimer is present again. If there is an internal link above to this one, I'm not sure so many references are necessary.  My 2c

MichaelBaj (changes requested):

docs/intro_installation_bootable_media.md
```diff
@@ -22,21 +23,29 @@ After imaging the ISO onto removable media, insert it into the target machine an
...
+
+To perform the installation using the OTP Installation, use the arrow keys to select either `OTP Install 128T Routing Software Serial Console` or `OTP Install 128T Routing Software VGA Console`. The OTP Installation for the serial or VGA console is described in the section [Router Installation Using OTP](intro_otp_iso_install.mdx).
 
 :::note
-Not all hardware has video support, therefore booting to console is the default (a console user may not be able to select an option). The default option is selected after a 30 second timeout.
+Not all hardware has video support. Booting to the serial console is the default, and is automatically selected after 30 seconds. When using the serial console, the terminal size is 80x25 - anything smaller may result in abnormal navigation behavior.
+
+Selecting the wrong type of console (Serial or VGA) may result in garbage characters being displayed, and if left to continue will result in an incorrect installation. If the wrong console is selected, reboot the target system and select the correct line for the target hardware.
```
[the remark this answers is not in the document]
jeffmart: corrected

docs/intro_installation_bootable_media.md
jeffmart: removed

docs/intro_installation_bootable_media.md
```diff
@@ -22,21 +23,29 @@ After imaging the ISO onto removable media, insert it into the target machine an
...
+
 #### 128T System via Serial Console
 
-Use this option when running on hardware with no video chipset. It uses `/dev/ttyS0` as the serial console for interacting with the installer.
+Use this option when running on hardware with no video chipset. It uses `/dev/ttyS0` as the serial console for interacting with the installer. 
 
-:::note
-Selecting the wrong type of console may result in garbage characters being displayed and the install hanging. If this is the case, reboot the target system and select the correct line for the target hardware.
-::: 
+The expected Serial console default size is 80x25. If the terminal being used to connect to the console is less than row (80) X column (25), the display may appear that the user is unable to select the entries, or the display does not repaint correctly. For serial console issues please refer to [Serial Console Troubleshooting](#serial-console-troubleshooting).
```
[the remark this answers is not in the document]
jeffmart: removed

MichaelBaj (approved):

PR #218, opened 2021-04-08
Add docs for monitoring agent plugin

agrawalkaushik:
- The monitoring agent document has been updated to capture the new plugin based installation and management

DOCS-209 #time 1d

### DIFF
--- a/docs/plugin_monitoring_agent.md
+++ b/docs/plugin_monitoring_agent.md
@@ -1,6 +1,6 @@
 ---
 title: 128T Monitoring Agent
-sidebar_label: Monitoring
+sidebar_label: Monitoring Agent
 ---
 
 The 128T Monitoring Agent is an entity for collecting data from a node running 128T software and to push it to a collector. It is capable of collecting the data from several sources such as metrics, events etc. The current mechanism of monitoring a 128T router involves performing REST or GraphQL queries from the conductor. At scale, this can become inefficient and be problematic in terms of the performance of the conductor. Additionally it is important to interact with 3rd party monitoring platforms as means for organizations to collect, analyze and report using various KPIs available from 128T software and other application in the network.
@@ -18,6 +18,14 @@ The monitoring agent at its core is designed to be able to push data to external
 
 ## Installation
 
+### Plugin
+For deployments running 128T version `5.1.0` or greater on the conductor should install the monitoring agent plugin for configuration management.
+
+:::note
+The instructions for installing and managing the plugin can be found [here](plugin_intro.md#installation-and-management).
+:::
+
+### Manual installation
 The 128T Monitoring Agent can be obtained from the official 128T software repository. The latest 128T-monitoring-agent should always be used and is compatible with `128T >= 4.1.0`
 
 :::important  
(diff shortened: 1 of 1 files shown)


docs/plugin_monitoring_agent.md
```diff
@@ -242,9 +469,48 @@ Path: `/var/lib/128t-monitoring/outputs/file.conf`
 
 #### Kafka
 
+The `kafka` output is one of the built in available types for the monitoring agent plugin. The various configuration options available under `authority > monitoring > output > kafka` are as follows
+
+| Element | Type        | Description                                                 |
+| ------- | ----------- | ----------------------------------------------------------- |
+| broker | list | List of kafka broker to communicate with                            |
```
gregschrock, line 476: kafka broker`s` to ...

docs/plugin_monitoring_agent.md
```diff
@@ -480,7 +901,45 @@ A custom set of metrics can be collected by configuring the `t128_metrics` input
...
 
-The event collector can be used for collecting and pushing events for various categories such as admin, alarm, system, traffic and provisioning as they occur on the system. The type of the event is available via a `tag` and can be used for filtering only specific events as desired. For example, the following configuration can be used for pushing just the `alarm` and `admin` event
+The event collector can be used for collecting and pushing events for various categories such as admin, alarm, system, traffic and provisioning as they occur on the system. The type of the event is available via a `tag` and can be used for filtering only specific events as desired. For example, the following configuration can be used for pushing just the `alarm` and `admin` event. The various configuration options available under `authority > monitoring > input > event` are as follows
+
+| Element | Type   | Default  | Description                                                  |
+| ------- | ------ | -------  | ------------------------------------------------------------ |
+| admin | boolean  | true     | Include admin events generated by the system                 |
```
gregschrock, line 908: We should probably note that these depend on the events being enabled (but not necessarily persisted) on the router itself.

docs/plugin_monitoring_agent.md
```diff
@@ -480,7 +901,45 @@ A custom set of metrics can be collected by configuring the `t128_metrics` input
...
+| audit | boolean  | true     | Include audit events generated by the system                 |
+| alarm | boolean  | true     | Include alarm events generated by the system                 |
+| traffic | boolean  | true     | Include traffic events generated by the system             |
+| provisioning | boolean  | true     | Include provisioning events generated by the system   |
+| system | boolean  | true     | Include system events generated by the system               |
+| track-index | boolean | true | Track event indexes such that it allows each event to be produced once when monitoring is restarted |
```
gregschrock, line 914: "produced once when monitoring" reads a bit odd. After a second read, I understand and I'm not sure how to improve it. Just thought I'd point it out in case others have suggestions.

docs/plugin_monitoring_agent.md
```diff
@@ -672,8 +1223,92 @@ The `lteMetricCollector128t` collector when run will scan the current node confi
 
 ### Top Analytics Collector
 
-The `topAnalyticsCollector128t` collector can be used for monitoring the top sources, top sessions and top applications on the router. The different aspects of each of these data sources are easily tunable using the input configuration.
+The `topAnalyticsCollector128t` collector can be used for monitoring the top sources, top sessions and top applications on the router. The different aspects of each of these data sources are easily tunable using the input configuration. The monitoring agent configuration provides a way to manage each of the top-sessions, top-sources and top-applications individually as follows
+
+#### Top Sessions
+The input type of `top-sessions` can be used to enable the top-sessions configuration. An example of such configuration is as follows
```
gregschrock, line 1229: This may be a reference thing, but it bothers me that there are a bunch of full sentences that end in `follows` with no punctuation. They're full, valid sentences. So they should end in a period; or a colon could be used to transition to the example.

docs/plugin_monitoring_agent.md
```diff
@@ -672,8 +1223,92 @@ The `lteMetricCollector128t` collector when run will scan the current node confi
...
+```
+
+#### Top Applications
+The monitoring agent `top-applications` input can be used to configure various aspects of the API on the system. The various configuration options available under `authority > monitoring > input > top-applications` are as follows
+
+| Element | Type    | Default | Description                                                 |
+| ------- | ------- | ------- | ----------------------------------------------------------- |
+| max-rows | uint32 | 10      | The maximum number of rows to be collected per sample       |
+| min-session-count | uint32 | 1 | The minimum number of sessions for an application to be collected |
+| application-filter | string | - | Filter for only including specific applications in the collection |
```
gregschrock, line 1286: It took me a couple tries to understand this filter description. Maybe `Restrict the applications to be included in the collection`?

docs/plugin_monitoring_agent.md
```diff
@@ -705,9 +1340,62 @@ The **top applications** input is useful when application identification in term
...
 | 3.3.1   | `t128_graphql` input type was introduced |
 
-The `t128_graphql` input can be used to retrieve data from a GraphQL API.
+The `t128_graphql` input can be used to retrieve data from a GraphQL API. The various configuration options available under `authority > monitoring > input > graphql` are as follows
 
+| Element | Type    | Description                                                 |
+| ------- | ------- | ----------------------------------------------------------- |
+| query-entry-point | string | The path to a point in the graphQL tree from which fields and tags will be extracted. This path may contain (`<key>:<value>`) graphQL arguments such as (name:'${ROUTER}'). | 
+| extract-field | list | List of leaf nodes to be collected from query response as fields |  
+| extract-field > name | string | The name of the collection |
```
gregschrock, line 1349: I know it's a reused grouping, but here I'd call out the specific `The name of the field` and `The name of the tag` rather than using the generic `collection`.

docs/plugin_monitoring_agent.md
```diff
@@ -757,7 +1445,18 @@ It is possible for the relative paths of `extract_fields` and `extract_paths` to
 
 When dealing with multiple child nodes, it is advised that each be handled in separate `t128_graphql` inputs.
 
-## Release Notes
+## Monitoring Agent Release Notes
```
gregschrock, line 1448: Are we going to have separate plugin and agent release notes sections?

docs/plugin_monitoring_agent.md
```diff
@@ -422,7 +794,56 @@ The 128T monitoring-agent comes pre-packaged with a set of collectors to assist
...
+                        name  internet-rx-bandwidth
+                        id    stats/aggregate-session/service/bandwidth-received
+                    exit
+
+                    metric              internet-tx-bandwidth stats/aggregate-session/service/bandwidth-received
+                        name  internet-tx-bandwidth
+                        id    stats/aggregate-session/service/bandwidth-received
+                    exit
+
+                    filter              service
```
gregschrock, line 835: Do we provide a way to itemize by a parameter without filtering it? I hadn't thought about that till now.

gregschrock (changes requested):
Just a couple minor things.

docs/plugin_monitoring_agent.md
```diff
@@ -18,6 +18,14 @@ The monitoring agent at its core is designed to be able to push data to external
 
 ## Installation
 
+### Plugin
+For deployments running 128T version `5.1.0` or greater on the conductor should install the monitoring agent plugin for configuration management.
```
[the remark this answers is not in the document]
agrawalkaushik: not sure I see the nuance of your comment. This is a plugin which needs to be installed on top of the existing 5.1.0 software. If you share a full line of what you intend this to read like it will be easier.

docs/plugin_monitoring_agent.md
agrawalkaushik: this is lifted from the yang description - I'm open to other suggestions.

docs/plugin_monitoring_agent.md
agrawalkaushik: Its not a collection filter but thats probably a subtle nuance we dont care about. I'll see what I can re-word it.

docs/plugin_monitoring_agent.md
agrawalkaushik: Eventually yeah, that was the idea - hence the reword. We typically don't document the "first" version but I can add a blurb here for clarity.

BenMatase (approved):

docs/plugin_monitoring_agent.md
gregschrock: No need for a reword. I just wanted to understand.

docs/plugin_monitoring_agent.md
agrawalkaushik: itemization is our default for the metrics collector https://github.com/128technology/telegraf/blob/06288994995f8017ed8be7e8d8a9bd020951bbcc/plugins/inputs/t128_metrics/retriever.go#L215

docs/plugin_monitoring_agent.md
agrawalkaushik: how about `Enable best effort tracking of events generated while the output cannot be reached`

docs/plugin_monitoring_agent.md
agrawalkaushik: done

BenMatase (approved):

MichaelBaj (approved):

PR #226, opened 2021-04-20
Write procedure for troubleshooting duplicate asset IDs

tcarroll25:
Write procedure for troubleshooting duplicate asset IDs and add that procedure to the troubleshooting section

### DIFF
--- a/docs/ts_ap_duplicate_assets.md
+++ b/docs/ts_ap_duplicate_assets.md
@@ -23,7 +23,7 @@ The `salt-key` command referenced above can be accessed from the Linux shell wit
 
 ## Symptoms
 
-An asset whose has a duplicate asset ID error can be seen from the conductor CLI using the `show assets` command with a provided asset-id. For example:
+An asset that has a duplicate asset ID error can be seen from the conductor CLI using the `show assets` command with a provided asset-id. For example:
 ```
 admin@node1.conductor# show assets asset1
 Tue 2021-04-20 14:21:30 UTC
--- a/docs/ts_ap_duplicate_assets.md
+++ b/docs/ts_ap_duplicate_assets.md
@@ -8,9 +8,9 @@ sidebar_label: "AP: Duplicate Asset ID Error"
 This document describes the process for troubleshooting the duplicate asset ID error with Automated Provisioning (AP).
 
 Here are a few situations that may cause a duplicate asset ID:
-* 128T SSRs being cloned from staging and have not properly completed their initialization to obtain a unique ID
-* The same quickstart file being applied to multiple 128T SSRs
-* A user manually entering the same asset ID into the salt-minion ID file on multiple 128T SSRs
+* 128T SSR has been cloned from staging and has not completed the initialization procedure, creating a unique ID
+* The same quickstart file has been applied to multiple 128T SSRs
+* A user has manually entered the same asset ID into the salt-minion ID file on multiple 128T SSRs
 
 :::note
 The terms "salt-master" and "conductor" are used interchangeably throughout this document. "Salt-master" refers to the salt-master process running on the conductor, which orchestrates tasks for AP.  Also the terms "minion", "salt-minion" and "asset" are used interchangeably throughout this document. "Minion" runs on an "asset", or system hosting a 128T router. Minions are responsible for carrying out tasks on the host, given to it by the salt-master.
--- a/docs/ts_ap_duplicate_assets.md
+++ b/docs/ts_ap_duplicate_assets.md
@@ -7,8 +7,13 @@ sidebar_label: "AP: Duplicate Asset ID Error"
 
 This document describes the process for troubleshooting the duplicate asset ID error with Automated Provisioning (AP).
 
+Here are a few situations that may cause a duplicate asset ID:
+* 128T SSRs being cloned from staging and have not properly completed their initialization to obtain a unique ID
+* The same quickstart file being applied to multiple 128T SSRs
+* A user manually entering the same asset ID into the salt-minion ID file on multiple 128T SSRs
+
 :::note
-The terms "salt-master" and "conductor" are used interchangeably throughout this document. "Salt-master" refers to the salt-master process running on conductor, which orchestrates tasks for AP.  Also the terms "minion", "salt-minion" and "asset" are used interchangeably throughout this document. "Minion" runs on an "asset", or system hosting a 128T router. Minions are responsible for carrying out tasks on the host, given to it by the salt-master.
+The terms "salt-master" and "conductor" are used interchangeably throughout this document. "Salt-master" refers to the salt-master process running on the conductor, which orchestrates tasks for AP.  Also the terms "minion", "salt-minion" and "asset" are used interchangeably throughout this document. "Minion" runs on an "asset", or system hosting a 128T router. Minions are responsible for carrying out tasks on the host, given to it by the salt-master.
 :::
 
 This document will reference salt keys and salt grains. Some good resources for these topics are:
@@ -78,13 +83,23 @@ asset4
 Rejected Keys:
 ```
 
-It is not always clear which situation is causing the duplicate asset ID error. If an asset is in the disconnected state and has a duplicate asset ID error then the conductor is unable to distinguish if the denied key is from the same salt-minion who is now disconnected, or the correct salt-minion is simply disconnected at the time when a new salt-minion tried to authenticate with a different public key and the same ID. However, if an asset is in any state where it is currently connected to the conductor (connected, installed, starting, running, stopped, etc.) and has a duplicate ID error then the user knows that they have a salt-minion with a duplicate ID trying to connect. But, they do not know if the asset currently connected is the correct asset, or if the asset with the denied key is the correct asset.
+:::note
+The `t128-salt` and `t128-salt-key` commands can only be executed by the `root` user at the Linux shell.
+:::
+
+There are multiple situations that may cause a duplicate asset ID error.
+
+1. A single salt-minion disconnected, had its key regenerated, and reconnected
+    * In this case, the asset will show as disconnected and have a duplicate asset ID error
+1. Two salt-minions with the same ID attempted to connect
+    * In this case, either the connected asset or the denied minion may be the correct one
+    * The correct asset may also be disconnected when another minion with the same ID connects, making it difficult to distinguish which situation caused the asset ID error
 
 ## Root Cause
 
 A denied key is a public key that is rejected automatically by the salt-master when the salt-minion tries to authenticate with its public key. A rejected key is a different state that indicates a user manually rejected a salt key. There are two different situations that can result in the salt-master denying a key:
 1. A salt-minion was rebuilt, reinstalled or the public/private key pair was wiped and new keys were automatically regenerated when the salt-minion service started. Now the minion's keys do not match the salt keys previously associated with the minion ID on the conductor.
-2. There are two salt-minions trying to authenticate with the same salt-minion ID.
+1. There are two salt-minions trying to authenticate with the same salt-minion ID.
 
 In the denied key state the salt-minion does not receive any communication from the salt-master. The conductor automatically creates a duplicate asset ID error when it detects that the salt-master has moved a salt key into the denied state.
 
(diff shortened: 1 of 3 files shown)


docs/ts_ap_duplicate_assets.md
```diff
@@ -0,0 +1,140 @@
...
+
+[Salt Grains](https://docs.saltproject.io/en/latest/ref/modules/all/salt.modules.grains.html)
+
+:::note
+The `salt-key` command referenced above can be accessed from the Linux shell with `t128-salt-key`. All arguments to the command and behavior are identical.
+:::
+
+## Symptoms
+
+An asset whose has a duplicate asset ID error can be seen from the conductor CLI using the `show assets` command with a provided asset-id. For example:
```
Chr1st0ph3rTurn3r, line 26: "An asset that has..."

Chr1st0ph3rTurn3r (approved):

docs/ts_ap_duplicate_assets.md
tcarroll25: Fixed

docs/ts_ap_duplicate_assets.md
```diff
@@ -0,0 +1,140 @@
...
+
+In the denied key state the salt-minion does not receive any communication from the salt-master. The conductor automatically creates a duplicate asset ID error when it detects that the salt-master has moved a salt key into the denied state.
+
+## Steps to Rectify
+
+There are two different procedures for rectifying the two different situations. If the user is unsure which situation they are in, then they should try the procedure for rectifying a salt-minion with new keys first, before moving onto locating multiple minions with the same ID.
+
+### Rectifying A Salt-Minion with New Keys
+
+The user needs to drop to the Linux shell and delete the accepted and denied keys with the same asset ID using the `t128-salt-key -d <asset-id>` command. For example:
```
jeffmart, line 97: While those that are familiar with Linux might recognize that the shell is for the root user (see #) might be good to call this out. I would not expect the t128-salt-key command to work as a non-root user.

docs/ts_ap_duplicate_assets.md
```diff
@@ -0,0 +1,140 @@
...
+title: "Automated Provisioner: Duplicate Asset ID Error"
+sidebar_label: "AP: Duplicate Asset ID Error"
+---
+
+## Introduction
+
+This document describes the process for troubleshooting the duplicate asset ID error with Automated Provisioning (AP).
+
+:::note
+The terms "salt-master" and "conductor" are used interchangeably throughout this document. "Salt-master" refers to the salt-master process running on conductor, which orchestrates tasks for AP.  Also the terms "minion", "salt-minion" and "asset" are used interchangeably throughout this document. "Minion" runs on an "asset", or system hosting a 128T router. Minions are responsible for carrying out tasks on the host, given to it by the salt-master.
```
jeffmart, line 11: nit: "process running on conductor" might flow better with "process running on the conductor"

jeffmart (changes requested):

docs/ts_ap_duplicate_assets.md
```diff
@@ -0,0 +1,140 @@
+---
+title: "Automated Provisioner: Duplicate Asset ID Error"
+sidebar_label: "AP: Duplicate Asset ID Error"
+---
+
+## Introduction
+
+This document describes the process for troubleshooting the duplicate asset ID error with Automated Provisioning (AP).
```
MichaelBaj, line 8: This introduction would benefit from an explanation of the situations that would result in a duplicate asset ID.
e.g., A duplicate Asset ID may be the cause of 128T SSRs being cloned from staging and have not properly completed their initialization to obtain a unique ID.

docs/ts_ap_duplicate_assets.md
```diff
@@ -0,0 +1,140 @@
...
+[root@conductor ~]#
+```
+
+:::note
+If the conductor is in HA configuration then the salt keys need to be deleted from *both* conductors or the asset will be stuck in the `key-accepted` state.
+:::
+
+The salt-minion may reach out again automatically and get accepted by the conductor if the asset ID matches an asset ID in the 128T configuration. If the salt-minion does not reach out then it needs to be restarted manually from the Linux command line with `systemctl restart salt-minion`.
+
+If the root cause was indeed a salt-minion that regenerated new keys and not two different salt-minions with the same ID, then this procedure will solve your problem and the asset will reconnect properly and no duplicate ID error will appear. If the duplicate asset ID error and the denied and accepted keys reappear then the user knows they are dealing with multiple salt-minions with the same minion ID.
```
MichaelBaj, line 127: Is this statement implying that it is possible for two minions to generate the same key? Shouldn't the odds of this be astronomically low? Have you seen this happen?

docs/ts_ap_duplicate_assets.md
```diff
@@ -0,0 +1,140 @@
...
+
+Tracking down the asset with the duplicate ID is not always easy. Since the salt-master automatically denies the authentication attempt from this bad actor, the conductor has no insight into the IP address or any other information about this asset. Secondly, the user does not know if the key that is currently denied belongs to the asset with the incorrect ID, or the correct one. Whichever system tries to authenticate first will get accepted and the other system will get denied. However, the user can use the salt command line to try and retrieve information from the asset who has been accepted to determine if this is the bad actor or the correct asset. For example, running the salt command `grains.items` will return a large amount of information about the system. Or the command `grains.get` can return specific pieces of information about the system. This might give the user a clue if the currently connected system has the correct minion ID. In the example below, the hostname grain retrieved was `asset10`, and if the customer is matching asset IDs to hostnames, then this may be a clue that this system is causing the duplicate ID error because the minion ID should be `asset10`:
```
MichaelBaj, line 133: This troubleshooting guide should point out the case where a bad actor may be attempting to connect to the Conductor by cloning an existing 128T SSR. I would add a section that goes through some validation steps to connect to the currently connected router to ensure its authenticity.

IMO, the right course of action is to deny both duplicate minions until the systems are authenticated. Once authenticated, regenerate a new minion ID and keys for the authentic minion and keep the other system denied until it can be claimed.

MichaelBaj (changes requested):

docs/ts_ap_duplicate_assets.md
```diff
@@ -0,0 +1,140 @@
...
+asset2
+asset3
+Denied Keys:
+asset1
+Unaccepted Keys:
+asset4
+Rejected Keys:
+```
+
+It is not always clear which situation is causing the duplicate asset ID error. If an asset is in the disconnected state and has a duplicate asset ID error then the conductor is unable to distinguish if the denied key is from the same salt-minion who is now disconnected, or the correct salt-minion is simply disconnected at the time when a new salt-minion tried to authenticate with a different public key and the same ID. However, if an asset is in any state where it is currently connected to the conductor (connected, installed, starting, running, stopped, etc.) and has a duplicate ID error then the user knows that they have a salt-minion with a duplicate ID trying to connect. But, they do not know if the asset currently connected is the correct asset, or if the asset with the denied key is the correct asset.
```
ian-h-chamberlain, line 81: This paragraph was a bit tough to follow for me, I might just simplify it to say something like


> There are multiple situations that may cause a duplicate asset ID error.
> 
> * A single salt-minion disconnected, had its key regenerated, and reconnected
>   * In this case, the asset may show as `disconnected`
> * Two salt-minions with the same ID attempted to connect
>   * In this case, either the `connected` asset or the denied minion may be the correct one

ian-h-chamberlain (approved):

docs/ts_ap_duplicate_assets.md
tcarroll25: I copy and pasted this from another doc, ill fix it in both places

docs/ts_ap_duplicate_assets.md
tcarroll25: Added a note at the first instance of `t128-salt-key` in the document

docs/ts_ap_duplicate_assets.md
tcarroll25: Good call, fixed

docs/ts_ap_duplicate_assets.md
tcarroll25: Good call, added that example and a few more

docs/ts_ap_duplicate_assets.md
tcarroll25: No thats not what I meant, I will clarify it. I meant two minions with the same ID and different keys.

docs/ts_ap_duplicate_assets.md
tcarroll25: @MichaelBaj `IMO, the right course of action is to deny both duplicate minions`: The only problem here is then we can't learn anything about either system. I think most people who want this document dont know where the hell all these duplicate asset IDs are and want help finding them, *cough* AAP. I'll try to add some more debugging options then end with delete both keys and hope you find the systems.

docs/ts_ap_duplicate_assets.md
```diff
@@ -7,8 +7,13 @@ sidebar_label: "AP: Duplicate Asset ID Error"
 
 This document describes the process for troubleshooting the duplicate asset ID error with Automated Provisioning (AP).
 
+Here are a few situations that may cause a duplicate asset ID:
+* 128T SSRs being cloned from staging and have not properly completed their initialization to obtain a unique ID
```
MichaelBaj, line 11: Minor rewording suggestion:
* 128T SSR has been cloned from staging and has not completed the initialization procedure, creating a unique ID

docs/ts_ap_duplicate_assets.md
```diff
@@ -78,13 +83,23 @@ asset4
...
+:::note
+The `t128-salt` and `t128-salt-key` commands can only be executed by the `root` user at the Linux shell.
+:::
+
+There are multiple situations that may cause a duplicate asset ID error.
+
+1. A single salt-minion disconnected, had its key regenerated, and reconnected
+    * In this case, the asset will show as disconnected and have a duplicate asset ID error
+1. Two salt-minions with the same ID attempted to connect
```
MichaelBaj, line 94: This should be 2.

docs/ts_ap_duplicate_assets.md
```diff
@@ -78,13 +83,23 @@ asset4
...
+    * In this case, either the connected asset or the denied minion may be the correct one
+    * The correct asset may also be disconnected when another minion with the same ID connects, making it difficult to distinguish which situation caused the asset ID error
 
 ## Root Cause
 
 A denied key is a public key that is rejected automatically by the salt-master when the salt-minion tries to authenticate with its public key. A rejected key is a different state that indicates a user manually rejected a salt key. There are two different situations that can result in the salt-master denying a key:
 1. A salt-minion was rebuilt, reinstalled or the public/private key pair was wiped and new keys were automatically regenerated when the salt-minion service started. Now the minion's keys do not match the salt keys previously associated with the minion ID on the conductor.
-2. There are two salt-minions trying to authenticate with the same salt-minion ID.
+1. There are two salt-minions trying to authenticate with the same salt-minion ID.
```
MichaelBaj, line 102: This should be 2.

docs/ts_ap_duplicate_assets.md
```diff
@@ -124,17 +139,53 @@ If the conductor is in HA configuration then the salt keys need to be deleted fr
...
 [root@t227-dut1 ~]# t128-salt asset1 grains.get host
 asset1:
     asset10
 ```
 
-Unfortunately the user can only query information about the system with the key that is accepted. If the user thinks the system that is currently accepted has the correct minion ID, they could try deleting both accepted and denied keys again and hope that the bad actor connects first and gets accepted so they can query information about that system.
+Unfortunately the user can only query information about the system with the key that is accepted. If the user has validated that the currently connected system is the correct system, then they could try the following procedure to locate the bad actor:
+1. Login to the correct system's Linux shell directly and stop the salt-minion with `systemctl stop salt-minion`
+    * Only do this if the user can maintain connectivity to this system so they can start the salt-minion again after this situation is resolved
+1. Delete both the accepted and denied salt-keys on the conductor(s) with `t128-salt-key` as stated above
```
MichaelBaj, line 185: I believe markdown expects that numbered lists increment sequentially (https://www.markdownguide.org/basic-syntax#ordered-lists)

MichaelBaj (changes requested):

docs/ts_ap_duplicate_assets.md
tcarroll25: From the best practices doc: "The numbers don’t have to be in numerical order, but the list should start with the number one.". The second row shows the list rendered correctly even with all 1s. I like it because I don't have to manually reorder all the numbers if I add new steps, markdown will always render it properly:

Screen Shot 2021-04-22 at 10 47 10 AM

docs/ts_ap_duplicate_assets.md
tcarroll25: done

docs/ts_ap_duplicate_assets.md
MichaelBaj: image

awesome

docs/ts_ap_duplicate_assets.md
tcarroll25: Markdown still renders properly per later comment

docs/ts_ap_duplicate_assets.md
tcarroll25: Markdown still renders properly per later comment

MichaelBaj (approved):

ian-h-chamberlain (approved):

jeffmart (approved):

PR #234, opened 2021-05-04
Update release_notes_128t_4.5.md

majes-github:
Wrong issue I95-38525 was referenced


jeffmart (approved):
Not clear why the last line is showing as a change, there are no differences

MichaelBaj (approved):

PR #250, opened 2021-06-08
Added note on negative impact with a router having hyperthreading enabled

jeffmart:


docs/about_supported_platforms.md
```diff
@@ -138,6 +138,8 @@ These are the minimum platform specifications for running the 128T Networking Pl
 
 :::note
 Multi-socket platforms are not compatible with the 128T software when run as a router.
+
+If hyperthreading is enabled on a router, the packet routing performance will be negatively impacted.
```
MichaelBaj, line 142: I don't believe this is true. It depends on the platform. We hired a developer who focused exclusively on addressing any hyperthreading issues. While the gains will not be great, they should be there. I know we haven't extensively tested hyperthreading performance.

PR #255, opened 2021-06-17
Update docs for monitoring agent 3.6.1 release

salegria8:
# Description
Release 3.6.1 will include a few bug fixes and a few new features. See changes below for more info.


docs/plugin_monitoring_agent.md
```diff
@@ -1348,10 +1348,10 @@ The `t128_graphql` input can be used to retrieve data from a GraphQL API. The va
 | query-entry-point | string | The path to a point in the graphQL tree from which fields and tags will be extracted. This path may contain (`<key>:<value>`) graphQL arguments such as (name:'${ROUTER}'). | 
 | extract-field | list | List of leaf nodes to be collected from query response as fields |  
 | extract-field > name | string | The name of the field |
-| extract-field > value | string | The graphQL query path relative to the entry-point from which to extract the value | 
+| extract-field > value | string | The graphQL query path from which to extract the value. The path can be relative to the entry-point or absolute. If the path is abolute, it cannot diverge from the entry point path and must exclude graphQL arguments |
```
BenMatase, line 1351: typo `absolute` in here

docs/plugin_monitoring_agent.md
```diff
@@ -1348,10 +1348,10 @@ The `t128_graphql` input can be used to retrieve data from a GraphQL API. The va
...
 | extract-field | list | List of leaf nodes to be collected from query response as fields |  
 | extract-field > name | string | The name of the field |
-| extract-field > value | string | The graphQL query path relative to the entry-point from which to extract the value | 
+| extract-field > value | string | The graphQL query path from which to extract the value. The path can be relative to the entry-point or absolute. If the path is abolute, it cannot diverge from the entry point path and must exclude graphQL arguments | 
 | extract-tag | list | List of leaf nodes to be collected from query response as tags | 
 | extract-tag > name | string | The name of the tag |
-| extract-tag > value | string | The graphQL query path relative to the entry-point from which to extract the value | 
+| extract-tag > value | string | The graphQL query path from which to extract the value. The path can be relative to the entry-point or absolute. If the path is abolute, it cannot diverge from the entry point path and must exclude graphQL arguments |
```
BenMatase, line 1354: What does it mean to diverge? Because the example below feels a little bit like diverging since it goes into the name

BenMatase: The same `abolute` typo is here as well

docs/plugin_monitoring_agent.md
```diff
@@ -1435,10 +1447,10 @@ The `TOML` configuration for the GraphQL input can be seen below
   The path to a point in the graph relative to which `extract_fields` and `extract_tags` will be specified. This path may contain `(<key>:<value)` arguments corresponding to those in the GraphQL tree.
 
 - **extract_fields**  
-  Paths, relative to the `entry_point`, from which fields should be created. Each value MUST point to a leaf in the graph. The keys become the field names for the produced values. At least one field MUST be specified.
+  Paths, absolute or relative to the `entry_point`, from which fields should be created. Each value MUST point to a leaf in the graph. If the path is absolute, it MUST NOT diverge from the `entry_point` and MUST exclude graphQL argument. The keys become the field names for the produced values. At least one field MUST be specified.
```
BenMatase, line 1450: `MUST exclude graphQL arguments`?

BenMatase (changes requested):
Just the typos

docs/plugin_monitoring_agent.md
gregschrock: done

docs/plugin_monitoring_agent.md
gregschrock: fixed `abolute`, but I'm not exactly sure about the question of diverging...

docs/plugin_monitoring_agent.md
```diff
@@ -1373,6 +1373,16 @@ config
                         value  enabled
                     exit
 
+                    extract-field      interface-count
+                        name   interface-count
+                        value  allRouters/nodes/nodes/nodes/deviceInterfaces/totalCount
```
agrawalkaushik, line 1378: technically this should just be `totalCount`, correct? Are you showing it like this as an alternate example? If so, I think its a bit confusing and consider removing it.

agrawalkaushik (approved):

docs/plugin_monitoring_agent.md
salegria8: `totalCount` for the path? if the path is just `totalCount` the collector will expect a leaf at `allRouters/nodes/node/nodes/deviceInterfaces/nodes/totalCount` which isn't right because of the `nodes` before `totalCount`

agrawalkaushik: that's fair - I didn't spot the missing `nodes` in the parent path - good point!

BenMatase (changes requested):
Of course the one time I do comments as part of a review I forget to submit. I guess the typo is the only thing I'd like to see changed

docs/plugin_monitoring_agent.md
Chr1st0ph3rTurn3r: From Dictionary.com: (of a road, route, or line) separate from another route, especially a main one, and go in a different direction.

docs/plugin_monitoring_agent.md
```diff
@@ -1432,22 +1444,81 @@ The `TOML` configuration for the GraphQL input can be seen below
...
+[[processors.t128_transform]]
+    transform = "diff"
+    remove-original = true
+    expiration = 30s
+
+    [processors.t128_transform.fields]
+        "delta" = "counter"
+```
+
+For this diff case, the transform will compute the delta from a "counter" field. No delta can be computed from a single point, so it will wait till the second observed value to produce the diff. If no values of observed for more than thirty seconds, it will start computing again as if it hadn't yet seen any values.
```
Chr1st0ph3rTurn3r, line 1513: "till" is not really a word. It's a contraction of "until" usually written as 'til. However, it shouldn't be used here. My suggestion: ...so the transform waits until the second observed value...
Also, try not to use contractions in documentation; "hadn't". My suggestion: If no values are observed for more than thirty seconds, the transform will begin computing again.

docs/plugin_monitoring_agent.md
```diff
@@ -1432,22 +1444,81 @@ The `TOML` configuration for the GraphQL input can be seen below
...
+| previous_fields | dict[string:string] | Optionally include a field indicating the previous observed value. Specify as `previous-field = new-field` such that the previous indicates the prior value of `new-field`. |
+
+The processors operates on fields separately for series that passes through it. For example, there may be multiple metrics passing through with a `value` field. The processor will distinguish between them using the tags and measurement name to identify each.
+
+When there is no value to produce, the processor will exclude the field. If the field name matches its source field, that field is removed. This can be a way of dropping lines as Telegraf will remove a line that has no fields (it isn't a valid line). For example, when watching for state changes on lines that have a single field, the original line will be dropped. This can be used to turn a polling mechanism into a means of notification.
+
+Processors are not currently exposed explicitly in the plugin config, but they can be achieved through an input's [additional config](plugin_monitoring_agent.md#input-configuration). Here are some example of the `TOML` configuration.
```
Chr1st0ph3rTurn3r, line 1486: examples of... (plural)

docs/plugin_monitoring_agent.md
```diff
@@ -1432,22 +1444,81 @@ The `TOML` configuration for the GraphQL input can be seen below
...
+| expiration      | duration            | If no value has been seen for this amount of time, begin again as if it is the first value. For `state-change` retransmit the current state.                                |
+| remove-original | bool                | Remove the original field (**NOT LINE**) in the case that the field is renamed.                                                                                             |
+| fields          | dict[string:string] | A mapping of `new-field = existing-field` where the new field is the computed transform of the existing field. If the names are the same, the field will be replaced.       |
+| previous_fields | dict[string:string] | Optionally include a field indicating the previous observed value. Specify as `previous-field = new-field` such that the previous indicates the prior value of `new-field`. |
+
+The processors operates on fields separately for series that passes through it. For example, there may be multiple metrics passing through with a `value` field. The processor will distinguish between them using the tags and measurement name to identify each.
```
Chr1st0ph3rTurn3r, line 1482: This paragraph is a little confusing not because of the content, but the grammar. My suggestion:
The processor operates on each field separately. For example, multiple metrics may pass through with a `value`field. The processor will distinguish between the metrics using the tags and measurement name.

docs/plugin_monitoring_agent.md
```diff
@@ -1432,22 +1444,81 @@ The `TOML` configuration for the GraphQL input can be seen below
...
+
+When there is no value to produce, the processor will exclude the field. If the field name matches its source field, that field is removed. This can be a way of dropping lines as Telegraf will remove a line that has no fields (it isn't a valid line). For example, when watching for state changes on lines that have a single field, the original line will be dropped. This can be used to turn a polling mechanism into a means of notification.
```
Chr1st0ph3rTurn3r, line 1484: Suggested wording for clarity:
When there is no value to produce, the processor excludes the field. If the field name matches the source field, the field is removed. If a line is not valid, for example, if it has no fields, Telegraf will remove it, effectively dropping the line. When watching for state changes on lines that have a single field, the original line will be dropped. In this way, a polling mechanism can be turned into a means of notification.

Chr1st0ph3rTurn3r (dismissed):

docs/plugin_monitoring_agent.md
```diff
@@ -1432,22 +1444,81 @@ The `TOML` configuration for the GraphQL input can be seen below
...
 
+
+## 128T Processors
+
+### Transform Processor
+
+| Release | Modification                                   |
+| ------- | ---------------------------------------------- |
+| 3.3.1   | `t128_transform` was introduced                |
+| 3.5.0   | `state-change` transform option was introduced |
```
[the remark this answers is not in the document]
gregschrock: Just curious if you have an example of this. I find it cleaner to have each change on a separate line. Often I want to know what version has a feature more than I want to know what features are in a version. It's subtle but seems easier to me as separate lines.

MichaelBaj (dismissed):

docs/plugin_monitoring_agent.md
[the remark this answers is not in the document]
BenMatase: That is fine

BenMatase (approved):

docs/plugin_monitoring_agent.md
```diff
@@ -1335,6 +1336,36 @@ The **top sources** input can be used to capture the list of source IP addresses
 
 The **top applications** input is useful when application identification in terms of module or tls have been configured on the router. By default, all the discovered sessions will be reported by the input. The user can tune the collection by setting up a search filter in the form of **filter** or eliminate the applications that have some minimum number of sessions via **min-session-count**. The **max-rows** will limit the reporting to the first N rows. The collection can be turned off by setting **enabled** to be false.
 
+#### Session Records
+
+| Release | Modification                                     |
+| ------- | ------------------------------------------------ |
+| 3.6.1   | `t128_session_records` input type was introduced |
+
+The monitoring agent `session-records` input can be used to session records generated by the system.
```
MichaelBaj, line 1345: This sentence does not appear to be grammatically correct.

The monitoring agent `session-records` input can be used to generate session records by the system.

docs/plugin_monitoring_agent.md
```diff
@@ -1512,6 +1543,59 @@ Diff:
...
+	# invert = false
+
+	## Whether to ignore if any tag or field keys are missing.
+	# ignore_missing_keys = false
+
+  [processors.t128_pass.condition.tags]
+	# tag1 = ["value1", "value2"]
+	# tag2 = ["value3"]
+
+  ## Fields work the same was a fields and can be included in the same condition.
```
Chr1st0ph3rTurn3r, line 1584: Wording is not clear; "Fields work the same way and can be..."

Chr1st0ph3rTurn3r (changes requested):

docs/plugin_monitoring_agent.md
salegria8: fixed

docs/plugin_monitoring_agent.md
salegria8: fixed

Chr1st0ph3rTurn3r (approved):

gregschrock (approved):

BenMatase (approved):

MichaelBaj (approved):

PR #264, opened 2021-07-08
adding supported drivers info for initial review.

Chr1st0ph3rTurn3r:


docs/about_supported_drivers.md
```diff
@@ -0,0 +1,734 @@
...
+| | BCM57414 NetXtreme-E 10Gb/25Gb RDMA Ethernet Controller (16d7) | net_bnxt | 4.5.0 - 5.3.0 |
+| | BCM57416 NetXtreme-E Dual-Media 10G RDMA Ethernet Controller (16d8) | net_bnxt | 4.5.0 - 5.3.0 |
+| | BCM57417 NetXtreme-E 10GBASE-T RDMA Ethernet Controller (16d9) | net_bnxt | 4.5.0 - 5.3.0 |
+| | BCM57412 NetXtreme-E Ethernet Partition (16de) | net_bnxt | 4.5.0 - 5.3.0 |
+| | BCM57417 NetXtreme-E 10Gb/25Gb RDMA Ethernet Controller (16e2) | net_bnxt | 4.5.0 - 5.3.0 |
+| | BCM57416 NetXtreme-E 10Gb RDMA Ethernet Controller (16e3) | net_bnxt | 4.5.0 - 5.3.0 |
+| | BCM57414 NetXtreme-E Ethernet Partition (16ec) | net_bnxt | 4.5.0 - 5.3.0 |
+| | BCM57416 NetXtreme-E Ethernet Partition (16ee) | net_bnxt | 4.5.0 - 5.3.0 |
+
+### Driver: net_bnx2xvf
```
MichaelBaj, line 84: From here on down, it appears that you still have the old formatting.  These should all be removed.

MichaelBaj (changes requested):

docs/about_supported_drivers.md
```diff
@@ -0,0 +1,734 @@
...
+| | T540-50A4 Unified Wire Ethernet Controller (54a4) | All |
+| | T522-50A5 Unified Wire Ethernet Controller (54a5) | All |
+| | T522-50A6 Unified Wire Ethernet Controller (54a6) | All |
+| | T580-50A7 Unified Wire Ethernet Controller (54a7) | All |
+| | T580-50A8 Unified Wire Ethernet Controller (54a8) | All |
+| | T580-50A9 Unified Wire Ethernet Controller (54a9) | All |
+| | T580-50AA Unified Wire Ethernet Controller (54aa) | All |
+| | T520-50AB Unified Wire Ethernet Controller (54ab) | All |
+| | T540-50AC Unified Wire Ethernet Controller (54ac) | All |
+| | Unknown Device (54ad) | All |
```
reidstidolph, line 206: Is `Unknown Device (54ad)` meaningful information? To me it isn't, but not sure if the `54ad` identifier would be helpful to anyone. If it is not, perhaps we remove these unknown devices.

docs/about_supported_drivers.md
Chr1st0ph3rTurn3r: Totally reformatted.

docs/about_supported_drivers.md
Chr1st0ph3rTurn3r: removed unknown devices.

docs/about_supported_drivers.md
```diff
@@ -0,0 +1,505 @@
...
+
+| Vendor ID | Device ID | Driver | Supported SSR/128T Version |
+| --- | --- | --- | --- |
+| Amazon.com, Inc. (1d0f) | Elastic Network Adapter (ENA) (ec20) | net_ena | 4.5.0 - 5.3.0 | 
+
+## Aquantia Corp
+
+| Vendor ID | Device ID | Driver | Supported SSR/128T Version |
+| --- | --- | --- | --- |
+| Aquantia Corp. (1d6a) | AQC107 NBase-T/IEEE 802.3bz Ethernet Controller [AQtion] (d107) | net_atlantic | 4.5.0 - 5.3.0 |
```
MichaelBaj, line 53: Rather than having a range for supported 128T version, I suggest that you have a single value for 128T support added, this way we don't have to update this page with each release. In the off case where we drop support for a driver or NIC (this will happen), we can add a column to those rows indicating when support was dropped.

MichaelBaj (changes requested):

MichaelBaj (approved):

jeffmart (approved):

dfitzpatrick128 (approved):

PR #271, opened 2021-08-11
updating ISO locations, and some rearranging of the information 

Chr1st0ph3rTurn3r:


docs/intro_downloading_iso.md
```diff
@@ -21,7 +21,9 @@ With releases 4.5.6-1 and 5.0.0-1 v2 (and greater), we now provide a single ISO
 - 128T-4.5.6-1.el7.OTP.v1.x86_64.iso
 - 128T-5.0.0-1.el7.OTP.v2.x86_64.iso
 
-The 128T Software package is available from our public servers located at: `https://software.128technology.com/artifactory/list/generic-128t-legacy-isos-remote/` using the username and token provided to you. The username/token simplifies access and does not require the conversion process used for the certificate. With the latest release of the 128T Installer, both a certificate or username/token can be used for installation. However, to use a username/token for software access, Installer 3.0.0 is required. 
+### Installer Considerations
+
+The username/token simplifies access and does not require the conversion process used for the certificate. With release 2.7 of the **128T Installer**, both a certificate or username/token can be used for installation. However, if you are using only a username/token for software access, Installer 3.0.0 is required.
```
jeffmart, line 26: Token support was only added in installer 3.0.0 (2.7 knows nothing about it). That being said tokens can be used to access any 128T software version, however certs can only access 5.2 and below. Was this to convey the installer version support or the 128T version support?

docs/intro_downloading_iso.md
```diff
@@ -31,7 +33,11 @@ Juniper highly recommends upgrading your Conductor to 4.5.7, 5.0.1, or 5.1.1. Th
...
 ## Downloading an ISO
 
-128 Technology ISOs are available from our public servers located at: `https://software.128technology.com/artifactory/list/generic-128t-legacy-isos-remote/` to users with an active account.
+The 128T Software packages are available from our public servers using the username and token provided to you. 
+
+For versions prior to release 5.3, use: `https://software.128technology.com/artifactory/list/generic-128t-legacy-isos-remote/`. 
+
+For version 5.3 and later, use `https://software.128technology.com/artifactory/list/generic-128t-isos-release-local/`. 
 
 You will be prompted for your username and token to access the web page listing the software versions, and will be able to download directly from the page. If your client certificate or token is currently installed on your 128 Technology routers and conductors, you can download software updates from our yum servers or the web page.
```
jeffmart, line 43: Do we want to update the section below this with the heading of:
"Downloading from the Command Line" ?

jeffmart (changes requested):
Want to understand the first section better as I think this will need to change

docs/intro_downloading_iso.md
Chr1st0ph3rTurn3r: I was attempting to convey Installer version support. 
This whole section is getting very complicated. I was trying to tease out the different requirements for different versions and organize it a bit. 
Taken by itself as above, it is not as confusing as trying to read the doc in the guide. 
Is it version 3.0 of the installer the old text is referring to? I can't tell from the way we worded it, and knowing there were two quick releases of the installer, 2.7 and 3.0.

docs/intro_downloading_iso.md
Chr1st0ph3rTurn3r: I'm not sure. I guess the question is, how likely are the users to not read the information above it and just cut and paste the pre-5.3 location? It would basically be repeating the info above it.

jeffmart (approved):

MichaelBaj (approved):

PR #280, opened 2021-09-10
Adding HA Sync Redundancy Plugin documentation

laneshields:
Adding documentation to explain the functionality of the HA Sync Redundancy plugin, which allows users to add a redundant connection to the HA Sync team and also allows users to use the HA Sync connection for HA Fabric traffic.

### DIFF
--- a/docs/plugin_ha_sync_redundancy.md
+++ b/docs/plugin_ha_sync_redundancy.md
@@ -1,0 +1,227 @@
+---
+title: HA Sync Redundancy Plugin
+sidebar_label: HA Sync Redundancy Plugin
+---
+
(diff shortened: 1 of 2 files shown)


MichaelBaj (approved):
Given how useful this is, I can't help wonder why we don't include this by default.

agrawalkaushik:
> Given how useful this is, I can't help wonder why we don't include this by default.

@MichaelBaj do you mean include the plugin by default in the product? If so, I agree - with 5.4.0 we have the ability to do that (we tried that with WAN assurance). Will look to add this to the mix of pre-installed plugins.

agrawalkaushik (approved):

MichaelBaj:
> > Given how useful this is, I can't help wonder why we don't include this by default.
> 
> @MichaelBaj do you mean include the plugin by default in the product? If so, I agree - with 5.4.0 we have the ability to do that (we tried that with WAN assurance). Will look to add this to the mix of pre-installed plugins.

Yes, @agrawalkaushik I do mean including this plugin as part of the default set of packages included along with the 128T.

PR #294, opened 2021-11-07
Add doc for http probe plugin

agrawalkaushik:


ptimmons (dismissed):

BenMatase (dismissed):

docs/plugin_http_probe.md
```diff
@@ -0,0 +1,280 @@
...
+| --    | --      | --          | --                                      |
+| name  | string  | key         | The name of the http probe profile      |
+| url   | string  | max: 2048   | A valid http or https URL to be used for availability checking      |
+| probe-interval | uint32 | default: 10 | The duration (in seconds) of how often to perform a link test to the destination |
+| number-of-attempts | uint32 | default: 4 | The number of consecutive HTTP(s) requests to be sent within the probe-duration before deciding that destination is unreachable |
+| probe-duration | uint32 | default: 5 | The duration (in seconds) within which to reach the destination. Each attempt will be made in (probe-duration / number-of-attempts) interval |
+| valid-status-code | list | at least 1 value required | The list of valid status codes to be expected from the server |
+
+* Example:
+```config {9-14}
```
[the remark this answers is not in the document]
agrawalkaushik: the `{9-14}` will highlight those specific lines in the code block. You'll be able to see this better once its published 😉

docs/plugin_http_probe.md
```diff
@@ -0,0 +1,280 @@
...
+
+* Config Path: authority > router[name] > service-route[name] > reachability-detection
+
+* Config Fields:
+The following fields should be enabled for the probe based detection to work. The http-probe mechanism is based on the concept of health probes described [here](https://docs.128technology.com/docs/config_service_health/#health-probes). The configuration fields below are required for the system to react to the loadbalancer API calls that the plugin makes.
+
+| Name  | Type    | Value | Description                             |
+| --    | --      | --          | --                                      |
+| enforcement | boolean | true | Toggle the configuration to be enabled for the reachability enforcement to take effect. |
+| probe-type | enumeration | always | For probe based reachability detection to take effect the probe-type must be set to `always`. |
```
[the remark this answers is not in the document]
agrawalkaushik: This is 128T config and cannot be renamed.

docs/plugin_http_probe.md
```diff
@@ -0,0 +1,280 @@
+---
+title: HTTP Probe Reachability Detection Plugin
+sidebar_label: HTTP Probe Reachability Detection
+---
+The http-probe plugin is designed to leverage the reachability detection APIs that were introduced in the 128T router as of 5.2.0 version of software. More documentation on the core product feature can be found [here](config_service_health.md)
```
MichaelBaj, line 5: s/128T/SSR/g

docs/plugin_http_probe.md
```diff
@@ -0,0 +1,280 @@
...
+
+* Launch a set of HTTP(s) monitor services which will probe the configured HTTP or HTTPs URL at the configured frequency
+* Periodically update the load-balancer APIs with the up/down status on service-paths associated with the http-probes
+
+:::note
+For HTTPS probes, the client will allow self-signed certificates for inspecting the reachability status of the service.
+:::
+
+## Configuration Snippet
+The plugin leverages the existing 128T reachability detection and enforcement configuration within the service-route on the router.
```
MichaelBaj, line 22: s/128T/SSR/g

jeffmart (changes requested):
MInor change

docs/plugin_http_probe.md
```diff
@@ -0,0 +1,280 @@
...
+                valid-status-code  202
+                valid-status-code  200
+            exit
+        exit
+    exit
+exit
+```
+
+### Service route configuration
+Once the profile is created, the next step is to enable the reachability enforcement and probe detection for a non SVR service-route and reference the profile in that config.
```
MichaelBaj, line 60: Is it true that this plugin is only applicable for non-SVR service-routes? What happens if it is configured for a SVR service-route? If there is a constraint that is not checked via validation, we need to have a note in the documentation.

jeffmart:
um ... I put up a change and the github display provided it but seems it did not take... not sure why

docs/plugin_http_probe.md
```diff
@@ -0,0 +1,280 @@
...
+| probe > probe-type | enumeration | http-probe | The probe-type must be set to `http-probe` in order to leverage the HTTP(s) based probing |
+| probe > http-probe-profile | reference | - | Reference to a previously configured http-probe-profile on the router |
+
+:::warning
+The `reachability-detection > probe` configuration allows for multiple probes of various types to be configured. All the probes for a given service-route must be up for the route to be considered as up.
+:::
+
+
+### Probe reachability
+The router should have a distinct path for the probe traffic. It is recommended that the user create dedicated services and service routes for this purpose. At this time, the plugin does not generate any configuration to ensure the reachability of the probe traffic to the destination. The probe will be originated in the default linux environment. In case of in-band management, the probe traffic will be associated with the `_internal_` tenant so care must be taken to allow the tenant when creating such config. As a reference the following configuration represents a service and route to reach one of the HTTP probes in this document.
```
MichaelBaj, line 81: Please reference in the troubleshooting section for what needs to be checked if the services and service-routes are not properly configured.

jeffmart:
Ok, I cannot seem to submit through the review change process here, my primary item is we have 128T references that are not in the actual configuration per say, for example in the doc sections we state:

"128T router" 

Are we still using this or which these references be changed to SSR?

docs/plugin_http_probe.md
```diff
@@ -0,0 +1,280 @@
...
+exit
+
+admin@node1.conductor1#
+
+```
+
+Each service-route is designed to probe a unique URL for that servers and monitors the health of the service at the TCP socket level as well as the HTTP stack. When one of the servers cannot be reached or responds with a unsuccessful status code (e.g. 404, 504 etc) the service path is taken out of service.
+
+:::warning
+When all the service routes associated with the same service are down, the default system behavior is to operates in a `best-effort` mode in which the physical link and L2 connectivity is used to determine the health of the path. In this case, its possible that sessions are routed to paths that are down from a probe perspective. As soon as one of the paths comes back in service, the load balancer will start using that path for all subsequent new sessions. The `best-effort` flag can be set to false for the associated service-policy to disable this behavior.
```
MichaelBaj, line 203: its -> it's

MichaelBaj (changes requested):

docs/plugin_http_probe.md
agrawalkaushik: I believe there to be validation for the reachability-enforcement which guarantees that. I'll double check that. Is your ask to include the text for the validation error or something along those lines or just a caveat?

docs/plugin_http_probe.md
agrawalkaushik: The configuration is at the service-route. There are product validations in place. Is the ask to include all those errors - I think they should be covered by the reachability enforcement feature that was published in 5.2. I'll double check or reference to those.

docs/plugin_http_probe.md
agrawalkaushik: I've updated the keyword where possible since there are process names etc with 128T in the name.

docs/plugin_http_probe.md
MichaelBaj: Please add a caveat since this is covered by validation.

docs/plugin_http_probe.md
MichaelBaj: If validation exists, then additional documentation is not needed.

docs/plugin_http_probe.md
agrawalkaushik: So this is an "implicit" requirement but there is no validation. I currently don't have the cycles to test the SVR to see if this just works. So my proposal is that we keep this caveat in place and we can update the doc based on additional testing later. @MichaelBaj let me know if that works for you.

jeffmart (dismissed):

jeffmart (dismissed):

MichaelBaj (dismissed):

MichaelBaj (approved):

PR #320, opened 2022-01-18
Document reachability-detection probe-type

robert-patterson:
- add a step in the enforcement mode checklist for probe-type, which
  controls if probes are considered in the service health check

### DIFF
--- a/docs/config_service_health.md
+++ b/docs/config_service_health.md
@@ -81,7 +81,7 @@ It is easiest to use the Service Routes configured in Detection Mode and simply 
 4. Under Reachability Detection Settings, enable Reachability Detection Enforcement.
 5. Define the Enforcement Detection Window (time). The `detection-window` (default 5s) determines how often the stats will be aggregated in terms of min, max, and median.
 6. Define an Enforcement Hold-Down time. This determines how long the path stays down upon determining it is unusable.
-7. If probes will be used to determine service health, configure `probe-type` as `always`. Otherwise, configure `probe-type` as `disabled` to use only organic traffic as the basis of service health.
+7. To use a probe to determine service health, configure the `probe-type` as `always`. To use only organic traffic to determine service health, set the `probe-type` to `disabled`.
 8. If you have previously configured a Reachability Profile, select the appropriate profile. If there is no existing Reachability Detection Profile, use the procedure below.
 9. Scroll to Reachability Detection Enforcement Probes and click ADD.
 10. Enter a name for the probe and click SAVE.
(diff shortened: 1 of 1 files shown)


MichaelBaj (dismissed):

avinashbhat80 (approved):

docs/config_service_health.md
```diff
@@ -81,11 +81,12 @@ It is easiest to use the Service Routes configured in Detection Mode and simply
 4. Under Reachability Detection Settings, enable Reachability Detection Enforcement.
 5. Define the Enforcement Detection Window (time). The `detection-window` (default 5s) determines how often the stats will be aggregated in terms of min, max, and median.
 6. Define an Enforcement Hold-Down time. This determines how long the path stays down upon determining it is unusable.
-7. If you have previously configured a Reachability Profile, select the appropriate profile. If there is no existing Reachability Detection Profile, use the procedure below.
-8. Scroll to Reachability Detection Enforcement Probes and click ADD.
-9. Enter a name for the probe and click SAVE.
-10. In the Probe Type window, select a probe type from the drop-down. The default probe type is an ICMP probe.
-11. Click Validate, and then Commit.
+7. If probes will be used to determine service health, configure `probe-type` as `always`. Otherwise, configure `probe-type` as `disabled` to use only organic traffic as the basis of service health.
```
Chr1st0ph3rTurn3r, line 84: Two things with this. First I suggest changing to active voice:
To use a probe to determine service health, configure the `probe-type` as `always`. To use only organic traffic to determine service health, set the `probe-type` to `disabled`.
Second, based on the config example below, it looks like the `probe-type` is configured within the `reachability-detection` config. I think this step should go under step 11 ("In the probe type window, select a probe type from the drop down.")

Chr1st0ph3rTurn3r (changes requested):

docs/config_service_health.md
robert-patterson: There's two different fields named `probe-type`. One is under reachability-detection (enum of always or disabled). The other is under `Reachability Detection Enforcement Probes`, which is the drop-down.
This new step is adding the `probe-type` under reachability-detection, which is at the same level as `Enforcement Detection Window` (step 5) and `Enforcement Hold-Down time` (step 6), so I think it makes sense to have this new step as 7. That way the user can do all the `Reachability Detection Settings` level settings before moving on to `Reachability Detection Enforcement Probes` in steps 8-11.

Chr1st0ph3rTurn3r: thank you for explaining that. I'm good with the changes.

Chr1st0ph3rTurn3r (approved):

lliu5374 (approved):

MichaelBaj (approved):

PR #321, opened 2022-01-18
Add Marvis Powered Search Component

thedillonb:
Provides a search component that leverages the Marvis search engine.

**Do not merge until migration of this docs site to Juniper is complete otherwise this will break search on the github hosted site.**


docusaurus.config.js
```diff
@@ -67,4 +67,11 @@ module.exports = {
     path.resolve(__dirname, './src/components/adobe-analytics-plugin'),
     './src/plugins/release-notes-api.js',
   ],
+  customFields: {
+    search: {
+      orgId: '0f342292-b359-4adf-b34c-a24aad8d3433',
+      token: '???',
```
thedillonb, line 73: @MichaelBaj we'll need to fill this in before we merge this - or figure out a means to populate this without me checking in the token.

MichaelBaj (approved):

docusaurus.config.js
MichaelBaj: The token is part of a private organization that was created expressly for this purpose. Additionally, as you put it, CORS will prevent the use of this token for other purposes, so it should be safe, correct?

MichaelBaj (approved):

src/theme/SearchBar.js
```diff
@@ -104,17 +102,16 @@ function Modal({onClose}) {
       setIsLoading(true);
 
       fetch(
-        `https://api.mist.com/api/v1/labs/orgs/${searchOptions.orgId}/knowledge_search`,
+        `https://ijx9l93j28.execute-api.us-east-1.amazonaws.com/default/Func`,
```
MichaelBaj, line 105: is this what the public name looks like? Does this change? Or do we have to pay for a dns entry when we stand this up proper?

docusaurus.config.js
```diff
@@ -67,12 +67,4 @@ module.exports = {
     path.resolve(__dirname, './src/components/adobe-analytics-plugin'),
     './src/plugins/release-notes-api.js',
   ],
-  customFields: {
```
MichaelBaj: Do you think it is also a good idea to checkin the lambda function as well? Even though this repo won't be using it, we can have it under source control?

MichaelBaj (approved):

PR #325, opened 2022-01-24
updates per DOCS-280. Adding port and protocol info for SVR.

Chr1st0ph3rTurn3r:
Will send offline pdf of the doc to Rob Schumann and Adam Morris unless they miraculously appear in GitHub.

### DIFF
--- a/docs/concepts_machine_communication.md
+++ b/docs/concepts_machine_communication.md
@@ -5,10 +5,10 @@ sidebar_label: Machine Communication
 This document lists the different communication channels between nodes within a router, between peering routers, and between routers and their conductor. This is to provide:
 
 - A guide for port forwarding when deploying a conductor behind a firewall
-- A tool to help predict and manage the amount of ambient, management-plane bandwidth that 128T software will use during "steady state." (This is useful when deploying devices that have limited use and/or expensive WAN connections, such as LTE.)
+- A tool to help predict and manage the amount of ambient, management-plane bandwidth that SSR software will use during "steady state." (This is useful when deploying devices that have limited use and/or expensive WAN connections, such as LTE.)
 
 ## Connections
-Each running instance of 128T software (generically termed as a _node_) can exist in one of two _roles_: a _router_ or a _conductor_. Furthermore, two _nodes_ can be coupled together to form a highly available router or conductor. There are machine to machine (M2M) connections between the various topologies of nodes, which will be covered here. These include:
+Each running instance of SSR software (generically termed as a _node_) can exist in one of two _roles_: a _router_ or a _conductor_. Furthermore, two _nodes_ can be coupled together to form a highly available router or conductor. There are machine to machine (M2M) connections between the various topologies of nodes, which will be covered here. These include:
 
 - Node to node connectivity within a highly available router
 - Router to router connectivity between peering routers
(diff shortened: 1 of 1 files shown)


MichaelBaj (approved):

MichaelBaj (approved):

docs/concepts_machine_communication.md
```diff
@@ -83,15 +83,40 @@ The firewall detector function sends deliberately malformed, out-of-order, and m
...
+
+#### Protocol
+
+By default, native TCP sessions use TCP, and UDP is used for all other session types. However, if firewall detection is enabled and [triggered](#firewall_detector), all traffic destined for a peer transitions to UDP. 
+
+You can manually configure traffic to use UDP for transport by modifying the [`udp-transform`](config_reference_guide.md#udp-transform) sub-element under [network-interface > adjacency](config_reference_guide.md#adjacency) or [network-interface > neighborhood](config_reference_guide.md#neighborhood).
+
+#### Port Ranges
+
+The default range of ports used for configuring waypoints with SVR is 16,385 through 65,533.
```
laneshields, line 108: I believe this range is split in half where the first half is used as an origin port for SVR and the second half is used as the destination port for SVR. This is done to avoid potential collisions. Not sure if it bears mentioning this fact (please confirm with someone in Engineering if so)

docs/concepts_machine_communication.md
```diff
@@ -83,15 +83,40 @@ The firewall detector function sends deliberately malformed, out-of-order, and m
...
+
+If you need to limit the ports or port range the SSR uses for receiving traffic, a `port-range` can be configured under the `neighborhood`. This tells peer SSRs to constrain the destination port range used when communicating with another router. Note that when manually specifying a port range, port numbers 1025 to 16383 can also be used. 
+
+#### Example:
+Let's say you want to utilize UDP for transport, but do not want to open up all the default ports. To limit the number of ports open on your firewall, you choose to specify a port range of 10,000 to 12,000. With each new waypoint, another 1,000,000 active sessions are added. Even with a small port range selection you can easily support 40,000 active users at 25 active sessions per user.
```
laneshields, line 113: I find the claim "With each new waypoint, another 1,000,000 active sessions are added" to be dubious here.  If we want to get specific then we can say each individual port open allows for (whatever the number of origin SVR ports as I commented above) new sessions per peer 128T router. Or just many thousands.

laneshields: Also, that math is specific to the number of 1M, so I'd pull that last sentence.

laneshields (changes requested):
Please amend the specific sentences that claim 1M active sessions as this is dubious and in most cases I don't think the router even supports 1M sessions

docs/concepts_machine_communication.md
Chr1st0ph3rTurn3r: This is covered in the Concepts topic Waypoints and Waypoint Ports. There it says that 
"By default, each 128T will allocate waypoint ports for each of its interfaces using the default range of 16,385 through 65,533, for a total of 49,148 ports. When sending traffic to a peer, the 128T will allocate an even numbered waypoint for itself and an odd numbered waypoint for its peer. (We did this to avoid "glare conditions," the exceedingly unlikely chance that two devices would pick the exact same values in two directions simultaneously)."
I think adding the text: **When sending traffic to a peer, the 128T will allocate an even numbered waypoint for itself and an odd numbered waypoint for its peer.** is useful here.

docs/concepts_machine_communication.md
Chr1st0ph3rTurn3r: As a former co-worker (tech writer) used to say "When in doubt, vague it out". I think going with less specific claims here is a better idea.

laneshields-jnpr (approved):

docs/concepts_machine_communication.md
Chr1st0ph3rTurn3r: verified with Engineering that this is correct.

laneshields (approved):

MichaelBaj (approved):

PR #364, opened 2022-03-31
Toc docs reorg

Chr1st0ph3rTurn3r:


sidebars.js
```diff
@@ -2,183 +2,247 @@ module.exports = {
...
+      "config_basics",
+      "config_templates",
+      "config_RBAC",
+      "config_asset_connection_resiliency",
+      "config_bgp",
+      "config_ospf",
+      "config_audit_event",
+      "config_dhcp",
+      "config_dns_proxy",
+      "config_forward_error_correction",
```
MichaelBaj, line 121: You will notice that some of the sidebar labels for configuration have "configuring x y and z" and many are just "feature name". We should be consistent. My preference is that since they are already in the configuration category, they do not need the sidebar label "configuring ..."

sidebars.js
```diff
@@ -2,183 +2,247 @@ module.exports = {
...
+      "concepts_session_timer",
+      "concepts_waypoint_ports",
     ],
-    "Installer/Initializer Reference": [
-      "installer_cli_reference",
-      "installer_preferences",
-      "initializer_cli_reference",
-      "initializer_preferences",
+    "Glossary": [
+      "concepts_glossary",
```
MichaelBaj, line 245: Perhaps remove the glossary category and move this item into concepts, or vice versa.

sidebars.js
```diff
@@ -2,183 +2,247 @@ module.exports = {
...
+      "howto_config_PPPoE",
+      "config_service_health",
+      "config_snmp",
+      "config_snmp_metrics",
+      "config_static_hostname_mapping", 
+      "config_transport_encryption",   
+    ],
+    "Ethernet Over SVR": [
+      "concepts_EthOverSVR",
+      "config_EthoSVR",
```
MichaelBaj, line 136: I like that you have pulled some of these config items out into their own respective feature heading. Should they be duplicated in the configuration section as well so someone can find it in either location?

Additionally, should each of the features be in their own "Features" category and start a 3-nested deep section?

MichaelBaj (changes requested):

PR #392, opened 2022-05-24
Removed advertisement-interface and capture length

MichaelBaj:


Chr1st0ph3rTurn3r (dismissed):

avinashbhat80 (approved):

ayungelson (approved):

Joe-Swetz (approved):

docs/config_ha_vrrp.md
```diff
@@ -90,20 +85,11 @@ Activate VRRP on the `wan` and `lan` device interfaces of node1. By configuring
                     name               wan
                     type               ethernet
                     pci-address        0000:00:05.0
-                    capture-filter     len>0
-
-                    vrrp
```
avinashbhat80, line 95: Is this intentional? Don't we want to show an example for `Wan` side redundancy too?

Joe-Swetz: I believe we allow configuring VRRP on the wan side, but I'm not sure if this is a common use case?

MichaelBaj: VRRP requires a switch in front of it, correct? It is atypical for WAN interfaces to be configured with VRRP in production deployments. While technically possible, I would prefer our examples to follow best practices. I know of one instance where VRRP was incorrectly provisioned on the WAN without supporting infrastructure.
I do value your input however. Let me know if I am making a mistake by removing this example.

avinashbhat80: I understand, the audience here is different and it makes sense. It is a well tested functionality and may be this is not a right place to advertise.  Just want to make sure the info is available some where for those who may be interested in wan side support.

avinashbhat80 (dismissed):

Chr1st0ph3rTurn3r (dismissed):

Chr1st0ph3rTurn3r (approved):

Joe-Swetz (approved):

PR #401, opened 2022-06-13
6.0 features

Chr1st0ph3rTurn3r:


docs/concepts_ssr_idp.md
```diff
@@ -0,0 +1,71 @@
+---
```
agrawalkaushik, line 1: At a high level, there are additional (important) details around the core pinning behavior and the fact that the router has to be restarted after configuring IDP etc that we need to account for. I dont know whats the best place to include that but I would recommend doing it in one of the docs.

PR #414, opened 2022-07-19
update image based install process, move image based repo info to top…

Chr1st0ph3rTurn3r:
… of downloads page, rearrange sidebar, and rename file from intro_installation_fips to -image.


BenMatase (approved):

docs/install_onboard_hdware.md
```diff
@@ -23,42 +23,58 @@ Use this data to complete the onboarding process for your whitebox device.
```
agrawalkaushik, line 23: We should clarify what the definition of whitebox device is somwhere on this page. We probably also want to rename this to be onboarding whitebox hardware instead of just onbox_hdware

P.s: The file name has a type `hdware` -> `hardware`

PR #449, opened 2022-09-21
fix plugin_cloud_ha

Alex-arch7:
1　add solution type alicloud-vpc which is testing and using in China
2   add RAM user policy details for solution type alicloud-vpc 
3   add example output for solution type alicloud-vpc 
4   add example configuration for solution type alicloud-vpc

### DIFF
--- a/docs/plugin_cloud_ha.md
+++ b/docs/plugin_cloud_ha.md
@@ -15,6 +15,7 @@ The instructions for installing and managing the plugin can be found [here](plug
 | ------------------ | --------------- | -------------------- |
 | Azure VNET         | `azure-vnet`    | 2.0.0                |
 | Azure Loadbalancer | `azure-lb`      | 2.0.0                |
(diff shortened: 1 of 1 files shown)


docs/plugin_cloud_ha.md
```diff
@@ -906,6 +964,136 @@ config
 exit
 ```
 
+
+### Part of Example Configuration #2 for alicloud-vpc after auto generation.
+
+Below is an example of a node and service part, it is a running configuration for user.
+
+:::warning
```
Chr1st0ph3rTurn3r, line 972:
```suggestion
:::info
```

Chr1st0ph3rTurn3r (changes requested):

agrawalkaushik (approved):

Chr1st0ph3rTurn3r (approved):

MichaelBaj (approved):

PR #455, opened 2022-10-12
BFD for BGP echo mode information

laneshields:
Added context about BFD asynchronous mode vs BFD echo mode due to confusion encountered in the field


docs/config_bfd.md
```diff
@@ -178,7 +178,11 @@ Tue 2022-02-01 15:39:13 UTC
 
 ```
 
+## BFD Echo mode tests
```
Chr1st0ph3rTurn3r, line 181:
```suggestion
## BFD Echo Mode Tests
```

docs/config_bfd.md
```diff
@@ -178,7 +178,11 @@ Tue 2022-02-01 15:39:13 UTC
...
 
+This feature enables BFD asynchronous mode support. Many routers additionally use BFD echo mode to ensure a forwarding path is available. While the SSR does not support the sending of BFD echo mode tests to a non-SSR router, we can return the echo mode traffic to the originating router allowing it to pass its echo mode test. Unlike BFD asynchronous mode, BFD echo mode is not handled by a BFD process on the peer router. Instead, it is intended to be processed by the packet forwarding plane and returned to the originating router. In order to pass this echo test traffic, an SSR must be configured to allow this traffic just like any other forwarding traffic. This requires a tenant and a service as with all traffic forwarded by the SSR.
```
Chr1st0ph3rTurn3r, line 183:
```suggestion
This feature enables BFD asynchronous mode support. Many routers use BFD echo mode to ensure a forwarding path is available. While the SSR does not support the sending of BFD echo mode tests to a non-SSR router, we can return the echo mode traffic to the originating router allowing it to pass its echo mode test. Unlike BFD asynchronous mode, BFD echo mode is not handled by a BFD process on the peer router. Instead, it is intended to be processed by the packet forwarding plane and returned to the originating router. In order to pass this echo test traffic, an SSR must be configured to allow this traffic just like any other forwarding traffic. This requires a tenant and a service as with all traffic forwarded by the SSR.
```

Chr1st0ph3rTurn3r (changes requested):

MichaelBaj (approved):

Chr1st0ph3rTurn3r (approved):

MichaelBaj (approved):

PR #517, opened 2023-02-07
Added disclaimer to software support policy and modifying env

MichaelBaj:


Chr1st0ph3rTurn3r (changes requested):
Minor tweaks:

SSR Software is built on top of an open source enterprise Linux distribution. Leveraging open source software affords many benefits. Among these is a flexible and stable OS that natively supports a range of deployments from OEM hardware platforms, to hypervisors, to public cloud deployments. Using a supported, enterprise open source OS means that thousands of developers are monitoring millions of lines of code in the Linux kernel and other packages - finding flaws and developing fixes before vulnerabilities become problems. No other operating system on the planet is nearly as flexible as Linux. In fact, there’s not much you can’t do with Linux.

To that end, some of our customers may choose to use this flexibility to their advantage by making changes to the operating system, packages, daemons, or scripts on an SSR platform. While there are few restrictions to prevent network operators from making changes to the system environment, installing or removing packages, or any such changes are not guaranteed to be supported with future versions of SSR software. Changes to the environment that impact the performance or reliability of the SSR software will not be covered under our support agreement. Changes made to the environment are not guaranteed to survive a software upgrade, and may need to be reapplied post-upgrade. It is the responsibility of the customer to validate that any changes made to the SSR environment are vetted and tested by the customer in their own lab environment.

Chr1st0ph3rTurn3r (approved):

PR #519, opened 2023-02-14
Remove outdated HA sync, reuse HA Sync Plugin info

Chr1st0ph3rTurn3r:


MichaelBaj (approved):

tcarroll25 (dismissed):

docs/plugin_ha_sync_redundancy.mdx
```diff
@@ -0,0 +1,26 @@
+---
```
agrawalkaushik, line 1: not sure what the `.mdx` format is for but this plugin is no longer necessary so please feel free to deprecate and remove all the documentation related to this plugin.

MichaelBaj: `.mdx` file is a markdown file that support javascript. Note the use of the `import` directives in the file.
Kaushik, I think I am missing something significant here. Is the plugin no longer a plugin and now there is native support for HA config without the use of the plugin? Why would the plugin no longer be necessary?

agrawalkaushik: If my memory serves me right the plugin was used before we had team interfaces. I talked a bit more with Lane and it looks like there are some other alternative use cases at bunch of customers. So please disregard my comment above the plugin not being necessary.

docs/plugin_ha_sync_redundancy.mdx
```diff
@@ -0,0 +1,26 @@
...
+## Configuration Snippet
+
+### Redundant HA Sync Interface
+<RedundantHASync/> 
+
+## Use Cases
+<HASyncUseCases/>
+
+## Troubleshooting
+<HASyncTroubleshooting/>
```
agrawalkaushik, line 23:
```suggestion
<HASyncTroubleshoot/>
```

Without the change, the troubleshooting section is not getting rendered.

agrawalkaushik: side note - just like we catch broken links, I wonder if docusaurus will let us catch this!

agrawalkaushik (changes requested):
lgtm!

Just need to fix the troubleshooting link!

docs/plugin_ha_sync_redundancy.mdx
Chr1st0ph3rTurn3r: Nice catch Kaushik. Docusaurus did not catch that. Nor did I in my local build.

MichaelBaj (approved):

docs/plugin_ha_sync_redundancy.mdx
agrawalkaushik: I usually spin up a local build to review doc changes... My click on the Troubleshooting page didn't work 😄

PR #526, opened 2023-02-28
Stuck flow recovery

Chr1st0ph3rTurn3r:


docs/config_session_recovery.md
```diff
@@ -0,0 +1,41 @@
...
+sidebar_label: Session Recovery Detection
+---
+
+### Version History
+
+| Release | Modification |
+| --- | --- |
+| 6.1.0 | Feature introduced |
+
+In a Spoke and Hub topology, forward and reverse flows are exchanged between devices. However, in a case where there is a service interruption on a hub and the session no longer exists, the hub can no longer process the flows sent by the spoke. The hub sends a message to the spoke to enable metadata, which often is flagged as a security issue.
```
MichaelBaj, line 12: While it is true that most of our deployments are hub-and-spoke, this feature pertains to SVR, which would be for any topology that consists of two peering SSR routers running SVR.
I don't see this distinction as important. Looking for @ayungelson to comment.

MichaelBaj: "which often is flagged as a security issue" - This should be changed to "which may falsely indicate to that the SSR is listening and responding to wayports." (or something similar that doesn't immediate raise security concerns)

docs/config_session_recovery.md
```diff
@@ -0,0 +1,41 @@
...
+![Session Recovery Detection](/img/config_session_recovery.png)
+
+### Session Recovery Detection
+
+Session Recovery Detection has two modes:
+
+- `packet-based` (default mode)
+- `inactivity-based` This has a configurable `inactivity-timeout` with a default of 5 seconds.
+
+Because this feature uses a request/response mechanism, **all** SSR's must be upgraded before using/enabling this feature.
```
MichaelBaj, line 33: use :::important::: admonition.

docs/config_session_recovery.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Show Commands
+
+Info coming
```
MichaelBaj, line 37: - [ ] todo

docs/config_session_recovery.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Show Stats Commands
+
+Info coming
```
MichaelBaj, line 41: - [ ] todo

MichaelBaj (changes requested):

docs/config_reference_guide.md
```diff
@@ -2549,6 +2549,27 @@ This controls whether the SSR will attempt to detect when session-optimization i
...
+## session-recovery-detection
+
+#### Path
+
+authority > session-recovery-detection
+
+#### Description
+
+When enabled, the originating node will monitor activity on the reverse flow. If no activity is detected for the specified time, it will add an additional metadata attribute called `session-health-check` to the next forward flow. The `session-health-check` attribute is validated by the receiving node and if the flow on the hub does not exist, the hub will generate an `enable-metadata` message back to the spoke.
+
```
ayungelson, line 2561: it's not just reverse, it's opposite... and it's not hub/spoke...
```
When enabled, the originating node will monitor activity on the opposite flow. If no activity is detected for the specified time, it will add an additional metadata attribute called `session-health-check` to the next packet. The `session-health-check` attribute is validated by the receiving node and if the flow on the other node does not exist, it will generate an `enable-metadata` message back to the originator.
```

docs/config_session_recovery.md
ayungelson: Yes, it has nothing to do with hub/spoke, it just applies to SVR.

ayungelson: > "which often is flagged as a security issue" - This should be changed to "which may falsely indicate to that the SSR is listening and responding to wayports." (or something similar that doesn't immediate raise security concerns)

Yes - I completely agree - Mike's wording is much better.

ayungelson:
Looks good to me.

MichaelBaj (changes requested):

PR #543, opened 2023-04-03
Corrected OS tuning docs to remove NETCONF

MichaelBaj:


Chr1st0ph3rTurn3r (approved):

docs/bcp_conductor_deployment.md
```diff
@@ -86,14 +86,14 @@ When deploying conductor on the internet, Juniper recommends limiting access to
...
-4. Adding the "salt-master" and "netconf" services to the `conductor` zone
+3. Creating a firewalld **service** for the conductor's IPC connection, containing TCP port 930; e.g., `ipc`
+4. Adding the "salt-master" and "ipc" services to the `conductor` zone
 5. Set the public address to be in the `conductor` zone
 
 Juniper has developed some salt states to assist with these common configuration steps. Refer to our [public Github repo](https://github.com/128technology/salt-states/blob/master/setup-firewalld-t128-zone.sls) for some samples you can use to tune your system accordingly.
 
 :::note
->>> The `netconf` configuration is not applicable to version 5.3 and later. NETCONF controls have been replaced with REST API controls in versions 5.1 and later, with no loss of functionality.
+>>> The `ipc` configuration is not applicable to version 5.3 and later. NETCONF controls have been replaced with REST API controls in versions 5.1 and later, with no loss of functionality.
```
ecarson, line 96: Not sure if this change is correct? This section was intended to convey the obsoleting of netconf in lieu of REST.

MichaelBaj: the term `ipc` is just use to refer to the rule, not the fact that NETCONF is actually in use. The primary use for 930 is configuration, and the fact that it was encoded as NETCONF is an implementation detail. The reason I posted this change is because a SE was confused by the fact that NETCONF was is use on port 930, so I wanted to remove the ambiguity.

ecarson (approved):

PR #550, opened 2023-04-19
adding info to release considerations for cpuProperties-core issue.

migolnikov:


Chr1st0ph3rTurn3r (dismissed):

ecarson (approved):

MichaelBaj (dismissed):

docs/release_notes_128t_6.0.md
```diff
@@ -23,6 +23,27 @@ Before upgrading please review the [**Upgrade Considerations**](intro_upgrade_co
...
 ------
 - **Plugin Upgrades:** If you are running with plugins, updates are required for some plugins **before** upgrading the conductor to SSR version 5.4.0 or higher. Please review the [Plugin Configuration Generation Changes](intro_upgrade_considerations.md#plugin-configuration-generation-changes) for additional information.
+------
+- **I95-51007 cpuProperties-core value isolating cores:** *In SSR software versions 6.0.0 and greater*, the `cpuProperties-cores` setting in `/etc/128technology/local.init` is erroneously isolating cores on conductor nodes when set. This setting was previously ignored on the conductor, but while resolving an earlier issue with the installer and initializer that allocated CPU cores for traffic, that is no longer the case. There should be no traffic cores allocated to or isolated on the conductor.
```
migolnikov, line 27: People may not understand why this is a problem. I would explain that "isolating cores for packet forwarding" and somehow remind that conductor does not do any packet forwarding.

docs/release_notes_128t_6.0.md
```diff
@@ -23,6 +23,27 @@ Before upgrading please review the [**Upgrade Considerations**](intro_upgrade_co
...
+``` 
+should be changed to:
+```
+  "cpuProperties" : {
+    "cores" : 0
+  },
+```
+
+Note that only the relevant section of `local.init` is shown for clarity. All other settings should be left the same.
+The change should be made on both nodes of an HA system. If a conductor is already running 6.0.0 or later it will be necessary to `systemctlf restart 128T` on each node after making this change. If the modification is made prior to upgrade it is not necessary to restart 128T service as this will be performed during the upgrade.
```
migolnikov, line 44: I would further say that setting this to 0 on the conductor is safe on all release going all the way back to version 1.0, but explain that no need to restart on versions prior to 6.0.0

docs/release_notes_128t_6.0.md
Chr1st0ph3rTurn3r: Changes made, please review asap

docs/release_notes_128t_6.0.md
Chr1st0ph3rTurn3r: changes made, please review asap.

migolnikov (commented):
ship it

PR #574, opened 2023-06-14
IDP improvements 6.1.4 r2

Chr1st0ph3rTurn3r:


docs/config_idp.md
```diff
@@ -99,4 +99,38 @@ config
...
+- Client IP Address prefix
+- Destination IP Address
+- Vulnerability 
+- Severity
+- Action Options 
+
+![New Rule Settings](/img/idp-profiles-rules.png)
+After creating the custom ruleset, use the `access-policy` as shown in the [Configuration section above](#configuration) to define the IDP policy where this custom ruleset is used. 
 
+**still need PCLI example of configuration, for consistency**
```
MichaelBaj, line 136: - [ ] TODO: Get PCLI output

docs/config_idp.md
```diff
@@ -99,4 +99,38 @@ config
 
 The IDP engine runs on each node of the system. Traffic is always sent to the IDP instance on the first node - per the order in configuration. Upon failover, the existing sessions do not gracefully failover; the TCP sessions are reset, and new sessions must be established by the client. For UDP sessions, the same best effort behavior can be expected from the IDP engine.
 
+## Custom IDP Policies and Rules
+
+The following is an example workflow:
+- An IDP policy is put in place.
+- Security events are triggered because traffic typical for that network configuration is considered a risk.
```
MichaelBaj, line 106: An example screen capture here would be illustrative

docs/config_idp.md
```diff
@@ -99,4 +99,38 @@ config
...
+To populate the Rules fields shown below, use the information available on the Security Events screen or from the [`show idp-events`](concepts_ssr_idp.md#security-events-dashboard) command.
+:::
+- Client IP Address prefix
+- Destination IP Address
+- Vulnerability 
+- Severity
+- Action Options 
+
+![New Rule Settings](/img/idp-profiles-rules.png)
+After creating the custom ruleset, use the `access-policy` as shown in the [Configuration section above](#configuration) to define the IDP policy where this custom ruleset is used.
```
MichaelBaj, line 134: @agrawalkaushik after making these custom IDP rules changes, does it require a restart of the IDP engine? If so, that would incur some downtime for traffic to start honoring the new rules. If that is the case, that must be indicated here in the docs that this will happen.

docs/concepts_ssr_idp.md
```diff
@@ -78,6 +78,10 @@ For information about changing the settings, see [Application Identification Mod
 
 ![App ID Update Frequency](/img/idp_app-id-data-update.jpg)
 
+## IDP Custom Rules
+
+Starting with SSR 6.1.4-R2, users can customize an existing base IDP policy, adding new or exception-based rules. For example, consider a case where an IDP policy is put in place, but security events are triggered because traffic typical for that network is considered a risk. Using an existing IDP policy, you can modify the profile to allow the specific traffic to flow as expected within the network. Changes are implemented from the SSR GUI or PCLI, and validated as part of the existing configuration workflow.
```
agrawalkaushik, line 83: > adding new or exception-based

We cannot add new rules - everything is just exception based. 

```suggestion
Starting with SSR 6.1.4-R2, users can customize an existing base IDP policy, adding exception-based rules. For example, consider a case where an IDP policy is put in place, but security events are triggered because traffic typical for that network is considered a risk. Using an existing IDP policy, you can modify the profile to allow the specific traffic to flow as expected within the network. Changes are implemented from the SSR GUI or PCLI, and validated as part of the existing configuration workflow. 
```

docs/config_idp.md
agrawalkaushik: @jamesesilvia can you please provide an example to include here

docs/config_idp.md
agrawalkaushik: It does not require a restart to enable the new policy changes. It does take some time for the changes to apply but our bypass logic should keep traffic off this policy until the IDP engine is ready (and there is an alarm to pair it up with). 

I see that Chris covered the startup behavior for IDP; @jamesesilvia @Chr1st0ph3rTurn3r I think its a good idea to capture a similar blurb for config changes as well. It's more selective where it only "bypasses" that one service and not the whole thing.

agrawalkaushik (approved):
overall lgtm!

docs/config_idp.md
jamesesilvia: Yeah will get some snippets in a bit.

docs/ts_idp.md
```diff
@@ -20,6 +20,14 @@ Thu 2022-05-19 04:45:49 UTC
  Docker:     running
 
 ```
+## IDP Start-up
+
+Beginning with SSR version 6.1.4-R2, traffic destined for IDP will temporarily bypass the IDP engine until it completes the startup process. This will take place at initial start up of the device, upgrade, or restart. An alarm is triggered on the device to indicate this is taking place, and is visible using the `show idp events` command. Once the IDP engine is fully booted, traffic is appropriately routed through IDP. Earlier versions routed traffic to the IDP engine before it was fully engaged, resulting in the loss of traffic. With `bypass-enabled` set to `true` (this is the default) traffic temporarily bypasses the IDP.
```
hadrielk, line 25: > An alarm is triggered on the device to indicate this is taking place

seriously?!?

hadrielk: Generate an alarm if and only if the IDP engine does not come up after X amount of time?

I mean that's an actual alarm able event, right? Not "boot up".

Chr1st0ph3rTurn3r: My understanding of how it works is that when the ssr or idp engine starts or restarts, IDP traffic automatically bypasses the IDP engine until it fully boots. During that time and alarm is raised so that the user/admin knows that traffic is NOT going through the IDP. As part of the info provided, we tell the user what traffic is not going through the IDP, and when the IDP engine will be fully booted and passing traffic. All that seems pretty important to me. If you are responsible for your the network security wouldn't you want to know that part of your security configuration is temporarily suspended, and when to expect it back?

hadrielk: > Point being, there may be cases where it does not start at all

Yes, that's I said "Generate an alarm if and only if the IDP engine does not come up after X amount of time?"

Because that's an actual unexpected event that the user not only needs to know about, but needs to do something about/fix.

But anyway, I'm resolving this issue thread.

docs/config_idp.md
```diff
@@ -99,4 +99,153 @@ config
 
 The IDP engine runs on each node of the system. Traffic is always sent to the IDP instance on the first node - per the order in configuration. Upon failover, the existing sessions do not gracefully failover; the TCP sessions are reset, and new sessions must be established by the client. For UDP sessions, the same best effort behavior can be expected from the IDP engine.
 
+## Modifing IDP Policies
+
+Starting with SSR 6.1.4-R2, users can customize an existing IDP policy, creating exception-based rules. The following is an example workflow:
+
+- A standard IDP policy is put in place.
```
jamesesilvia, line 106: nit: wording `standard` is overloaded since its a possible `idp-policy` option and an adjective 😉 

change to: A "strict" `idp-policy` is initially configured for the service below.

docs/config_idp.md
```diff
@@ -99,4 +99,153 @@ config
...
+```
+admin@node.cadillac# show idp events since 2m
+Wed 2023-07-12 14:19:00 UTC
+========== ================================== ======== ============= =============================== ========== ========
+ Severity   Time                               Threat   Application   Attack                          Protocol   Action
+========== ================================== ======== ============= =============================== ========== ========
+        6   2023-07-12 14:17:54.659000+00:00   MEDIUM   HTTP          HTTP:EXTRA_CONTROL_CHARACTERS   TCP        CLOSE
+```
+
+- Use an existing IDP policy to create a custom IDP profile in the Authority Settings configuration that can be referenced in the `access-policy`.
```
jamesesilvia, line 133: In this case, the User has identified the attack of `HTTP:EXTRA_CONTROL_CHARACTERS` to be of minimal risk to this network and would like to create a rule to change this event to be `alert` only.

docs/config_idp.md
```diff
@@ -99,4 +99,153 @@ config
...
+admin@node.cadillac# show idp events since 2m
+Wed 2023-07-12 14:19:00 UTC
+========== ================================== ======== ============= =============================== ========== ========
+ Severity   Time                               Threat   Application   Attack                          Protocol   Action
+========== ================================== ======== ============= =============================== ========== ========
+        6   2023-07-12 14:17:54.659000+00:00   MEDIUM   HTTP          HTTP:EXTRA_CONTROL_CHARACTERS   TCP        CLOSE
+```
+
+- Use an existing IDP policy to create a custom IDP profile in the Authority Settings configuration that can be referenced in the `access-policy`. 
+- Modify the profile to allow the specific traffic to flow as expected within the network, for example, modifying the action to alert/allow. Changes are implemented from the SSR GUI or PCLI (in this example the PCLI is used), and validated as part of the existing configuration workflow.
```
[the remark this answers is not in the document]
jamesesilvia: Just flow/wording, No changes needed here

jamesesilvia (changes requested):

MichaelBaj (approved):

jamesesilvia (approved):

PR #576, opened 2023-06-16
Multicast 6.1.4 r2

Chr1st0ph3rTurn3r:


docs/config_multicast.md
```diff
@@ -0,0 +1,216 @@
...
+
+                        multihop
+                            ttl  5
+                        exit
+                    exit
+                exit
+```
+3. Add the routing protocol (IGMP) that defines an interface and IGMP version, as well as the IGMP Joined Group.
+```
+				    igmp
```
MichaelBaj, line 92: This `igmp` seems misplaced in the config. perhaps this is the end of the igmp config where you specify the version, per description #3.

docs/config_multicast.md
```diff
@@ -0,0 +1,216 @@
...
+### Multicast Configuration
+
+:::note
+BGP over SVR is a prerequisite for Multicast over SVR.
+:::
+
+The following components are necessary to configure multicast on the SSR. Configuration for each of the components is provided below, and a full example configuration is provided at the end of this process. 
+
+- Routing Interface: Specify the BGP over SVR interface
+- BGP over SVR: See [BGP over SVR](config_bgp.md#bgp-over-svr-bgposvr) for configuration details.
```
MichaelBaj, line 30: I don't recall from the functional specification why provisioning BGP is necessary for multicast. Can you point me to supporting documentation? This may be required for the customer as well.

docs/config_multicast.md
```diff
@@ -0,0 +1,216 @@
...
+The following components are necessary to configure multicast on the SSR. Configuration for each of the components is provided below, and a full example configuration is provided at the end of this process. 
+
+- Routing Interface: Specify the BGP over SVR interface
+- BGP over SVR: See [BGP over SVR](config_bgp.md#bgp-over-svr-bgposvr) for configuration details. 
+- Routing protocol such as IGMPv2/v3
+- PIM
+- RP
+- Tenant: defined as multicast. For additional info about Tenants, see [Tenants](config_tenants.md)
+- Service: Services are created for the multicast traffic, and defined for the Group address.
+- Access Policy: Allows receivers on the outgoing interfaces
```
robertkebler, line 36: Configures the tenants that are allowed to be receivers for the Multicast Service

docs/config_multicast.md
```diff
@@ -0,0 +1,217 @@
...
+In order to support Multicast over SVR, the BGP over SVR configuration is used in conjunction with the PIM configuration to generate auto-provisioned services and service-paths. BGP over SVR must be provisioned to support the Multicast feature. See [BGP over SVR](config_bgp.md#bgp-over-svr-bgposvr) for configuration details. 
+:::
+
+The following are the components necessary to configure multicast over SVR on the SSR. Configuration for each of the components is provided below, and a full example configuration is provided at the end of this process. 
+
+- Routing Interface: Specify the BGP over SVR interface
+- BGP over SVR: See [BGP over SVR](config_bgp.md#bgp-over-svr-bgposvr) for configuration details. 
+- Routing protocol such as IGMPv2/v3
+- PIM
+- RP
```
laneshields, line 34: It might be worth explicitly stating that RP is Rendevous Point. I think we should also specify that in the current implementation the RP needs to be external to the SSR fabric. We cannot support configuring the RP on the SSR itself.

docs/cli_reference.md
```diff
@@ -3,7 +3,7 @@ title: Command Line Reference
 sidebar_label: Command Line Reference
 ---
 
-The Command Line Reference guide is better understood if you know the basics of operating the programmable command line interface (PCLI). Commands and actions such as clear, edit, delete, restore, and show, for example, are described here. If you have not used the PCLI before, please refer to [About the PCLI](concepts_pcli.md) for an explanation of how it works.
+The Command Line Reference guide is better understood if you know the basics of operating the programmable command line interface (PCLI). Commands and actions such as clear, edit, delete, restore, and show, for example, are described here. If you have not used the PCLI before, please refer to [About the PCLI](concepts_pcli.md) for an explanation of how it works.
```
hadrielk, line 6: this line change only appears to add an extra space at the end?

docs/config_multicast.md
```diff
@@ -0,0 +1,217 @@
...
+- Tenant: Defined as multicast. For additional info about Tenants, see [Tenants](config_tenants.md)
```
laneshields, line 35: I'm not sure what "defined as multicast" means for this. I think it means that in the config examples below, there is a tenant named multicast. But that's just for the below config example. And that's only one of the tenants at play because there is also a tenant customerA-mc that is required.

It's clear that you need tenants for this because you need tenants for us to pass any traffic. What I think needs to be stressed for multicast is that there are two "types" of tenants as configured in the service. There is the "access-policy" tenant(s) that are the tenants allowed to receive the multicast traffic via IGMP joins at routers in the network. There is also the "multicast-sender-policy" which defines the tenant that is allowed to be the source of the multicast stream. This adds a little color to the bullet points.

docs/config_multicast.md
```diff
@@ -0,0 +1,217 @@
...
+        exit
+```
+6. Create the Service. 
+```
+        service            customerA-mc
+            name                     customerA-mc
+            security                 aes1
+            address                  224.0.0.0/4
+```
+7. Add an Access policy: An access policy is not required, however if an access policy is not specifically configured, then all interfaces will allow "joins".
```
laneshields, line 183: ok, did not know this. That might change somewhat my comment above

docs/cli_reference.md
Chr1st0ph3rTurn3r: That's funny. Not something I did... but that's not the first time I've seen changes that I never touched. I think github must make random updates like that. I have no idea how something like that would even happen.

laneshields (approved):

docs/config_multicast.md
```diff
@@ -0,0 +1,299 @@
...
+- Tenants: Two types of tenants that are necessary; the access-policy tenant is the tenant allowed to receive the multicast traffic via IGMP joins at routers in the network. The multicast-sender-policy defines the tenant that is allowed to be the source of the multicast stream. For additional info about Tenants, see [Tenants](config_tenants.md)
+- Service: Services are created for the multicast traffic, and defined for the Group address.
+- Access Policy: Configures the tenants that are allowed to be receivers for the Multicast Service.
+- Multicast Sender Policy: Configures the tenants that are allowed for the Multicast service. 
+
+### Configuration Process
+
+The following diagram represents a simple multicast configuration. A Client sends a multicast to address 225.1.1.1. Location1 and Location2 want to receive this multicast stream. The network is setup to use an external, non-SSR Rendezvous Point (RP) at 20.20.20.20 for all multicast traffic. This router is 192.168.22.2 and is configured for BGP with SSR rtr2. It advertises 20.20.20.20 as the RP for multicast into BGP,  allowing the SSR network to identify the path to the RP.
```
robertkebler, line 42: "A Client sends a multicast to address 225.1.1.1"
I don't understand this statement.  I don't see 225.1.1.1 anywhere in the config below.  Should this just be removed?

"It advertises 20.20.20.20 as the RP for multicast into BGP"
The RP is manually configured on each router, and is not advertised.

robertkebler (approved):

docs/config_multicast.md
```diff
@@ -14,185 +14,268 @@ Please refer to the [Juniper Multicast Overview](https://www.juniper.net/documen
...
+The following are the components necessary to configure multicast over SVR on the SSR. Configuration for each of the components is provided below, and a full example configuration is provided at the end of this process. 
 
 - Routing Interface: Specify the BGP over SVR interface
 - BGP over SVR: See [BGP over SVR](config_bgp.md#bgp-over-svr-bgposvr) for configuration details. 
 - Routing protocol such as IGMPv2/v3
 - PIM
-- RP
-- Tenant: defined as multicast. For additional info about Tenants, see [Tenants](config_tenants.md)
+- RP: The Rendezvous Point. In the current implementation, the RP must be external to the SSR. 
+- Tenants: Two types of tenants that are necessary; the access-policy tenant is the tenant allowed to receive the multicast traffic via IGMP joins at routers in the network. The multicast-sender-policy defines the tenant that is allowed to be the source of the multicast stream. For additional info about Tenants, see [Tenants](config_tenants.md)
```
MichaelBaj, line 35: is this access-policy by way of tenancy redundant with the access policy statement below?

MichaelBaj (approved):

docs/config_multicast.md
laneshields: I think this came from the back of napkin drawing I gave. I just picked a random multicast address to use for an example stream. The service in the config (and the RP config) is 224/4, so it would be inclusive of that address.

PR #584, opened 2023-07-11
Changed feature name for bfd tunnel

reidstidolph-JNPR:
The feature name for the `bfd-tunnel` setting of `peer-path-overlay` is word salad. Better late than never, I'm proposing a more concise feature name that is also descriptive of the intended purpose of the setting. This will get the docs to reflect the naming we are wanting to use in NPI materials for the feature.

### DIFF
--- a/docs/config_bfd_tunnel.md
+++ b/docs/config_bfd_tunnel.md
@@ -1,9 +1,9 @@
 ---
-title: BFD Overlay over SVR
-sidebar_label: BFD Overlay over SVR
+title: SVR Transport Reuse
+sidebar_label: SVR Transport Reuse
 ---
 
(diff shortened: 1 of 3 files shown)


MichaelBaj (approved):

docs/config_bfd_tunnel.md
```diff
@@ -13,7 +13,7 @@ This feature is accessed from Authority > Router > Node > Device > Network Inter
 Use the path described above to access the Neighborhood configuration. After naming the neighborhood, choose the `spoke` topology type. This opens the Peer Path Overlay field. The following options are available:
 
 - **svr**: The original SVR overlay connection method. This is the default method for bidirectional connectivity. 
-- **bfd-tunnel**: Outbound only connection method; enables the BFD Overlay over SVR feature. 
+- **bfd-tunnel**: Outbound only connection method; enables the SVR Transport Reuse feature.
```
ayungelson, line 16: What am I missing - we can't just change the documentation - don't we need to also change the config option + all the user visible things such as metrics?

reidstidolph-JNPR: @ayungelson we change documentation all the time. Initial docs often need improved and amended to better convey a feature. No functionality or metric need changed here, just the words used to describe it.

PR #592, opened 2023-07-28
removing references to version 4.x software

Chr1st0ph3rTurn3r:


docs/releases.table.js
```diff
@@ -295,16 +295,6 @@ export const releases = [
     url: '/docs/release_notes_128t_5.1#release-514',
     releaseDate: 'July 7, 2021',
   },
-  {
```
MichaelBaj, line 298: Junos does not remove old release notes. As far as I can tell, they are kept indefinitely.
The point of this exercise is to remove incorrect or irrelevant references.

MichaelBaj (changes requested):

docs/releases.table.js
MichaelBaj: These still show as being removed from the sidebar.

MichaelBaj (changes requested):

MichaelBaj (approved):

PR #595, opened 2023-08-09
adding session optimization documentation

Chr1st0ph3rTurn3r:


docs/config_session_optimization.md
```diff
@@ -0,0 +1,15 @@
...
+TCP session optimization is specifically for customers using low-bandwidth, high-latency links, that experience random drops not due to traffic congestion. In this environment, session optimization will override congestion control and the `send` window to provide better performance for the TCP stream over low-bandwidth, high-latency, lossy links. 
+
+To identify paths for TCP session optimization, BFD must indicate that the path has a high latency (more than half a second). The following settings are then configured on the device-interface.  
+
+1. Set a [`traffic-engineering transmit-cap`](config_reference_guide.md#traffic-engineering) of 10 Mb or lower on the desired device-interface.
+2. The `enable-detection` field under [`device-interface > session-optimization`](config_reference_guide.md#session-optimization-device-interface) must be true (true is the default value).
+
+### Caveats
+
+Because the packets that are stored for this feature cannot be easily backed up and recovered `session-optimization` should not be configured on High Availability nodes, or inline with other WAN optimization features that employ flow move or route modify.
```
reidstidolph-JNPR, line 15: "...employ flow move or route modify." Can we put this in terms that a customer would understand? I'm not even sure what situations we are describing here, which should be avoided.

reidstidolph-JNPR (changes requested):
Requesting clarification on what "WAN optimization features that employ flow move or route modify" means, that a typical user without deep implementation knowledge would understand.

docs/config_session_optimization.md
Chr1st0ph3rTurn3r: the section has been revised with input from Don Troshynski.

docs/config_session_optimization.md
```diff
@@ -0,0 +1,19 @@
+---
+title: TCP Session Optimization
+sidebar_label: TCP Session Optimization
+---
+
+TCP session optimization is specifically for customers using low-bandwidth, high-latency links, that experience random drops not due to traffic congestion. In this environment, session optimization will override congestion control and the `send` window to provide better performance for the TCP stream over low-bandwidth, high-latency, lossy links. 
+
+To identify paths for TCP session optimization, BFD must indicate that the path has a high latency (more than half a second). The following settings are then configured on the device-interface.  
+
+1. Set a [`traffic-engineering transmit-cap`](config_reference_guide.md#traffic-engineering) of 10Mbs (or maybe 25Mbps - checking on this) or lower on the desired device-interface.
```
ayungelson, line 10: the limit in the code now is 30 Mbps

docs/config_session_optimization.md
```diff
@@ -0,0 +1,19 @@
...
+2. The `enable-detection` field under [`device-interface > session-optimization`](config_reference_guide.md#session-optimization-device-interface) must be set to `true` (true is the default value).
+
+### Limitations
+
+Session failover for optimized sessions is not supported in HA configurations.
```
ayungelson, line 15: I am 100% certain that there are more features that are not supported when session opt is enabled. The list of limitations is probably a mile long.  I am guessing that app-id would be severely affected. @svjaffer would be able to comment more. This feature is very limited, and hasn't really been battle tested for many years at this point and for every POC that tried to use it, it fell short.

MichaelBaj: Recommended proposal. Rather than listing all features that may not work, instead define how this feature is known to work.

Session failover is designed to work in non-HA configurations without other features enabled.

Obviously "other features enabled" is too broad. Need help with clarification.

MichaelBaj (changes requested):

PR #598, opened 2023-08-24
Vmware getting started doc

Chr1st0ph3rTurn3r:


docs/install_vmware_config.mdx
```diff
@@ -0,0 +1,149 @@
...
+title: Installing in VMWare
+sidebar_label: Installing in VMWare
+---
+
+import InitialBoot from './_install_nmtui_router.md'; 
+
+
+This guide describes the process for deploying a Session Smart Conductor or a Session Smart Router (SSR) in VMWare ESXi. The process consists of the following steps:
+
+- Download the SSR package-based or image-based ISO. Refer to [Downloading ISOs](intro_downloading_iso.md)
```
MichaelBaj, line 11: As of today I don't believe there is a use case where image-based deployments can work for ESXi since they will be conductor-managed.
@agrawalkaushik could an ESXi be managed as a whitebox in Mist?

docs/install_vmware_config.mdx
```diff
@@ -0,0 +1,149 @@
...
+- Log into VMware ESXi with Administrative rights
+- Create a new Virtual Machine and configure
+- Boot Virtual Machine to run the Juniper SSR ISO and install software
+
+## VMWare Configuration
+
+In order to perform the configuration and installation of the SSR software, you must have a properly installed and working configuration of VMware ESXi. This documentation references VMware ESXi version 7.0.0, however, the software has been tested on earlier versions. Additoinally, you must have administrator access to the ESXi platform.
```
MichaelBaj, line 18: Typo "Additoinally"

docs/install_vmware_config.mdx
MichaelBaj: Also, need to confirm from @drcosic if `qcow` should be used for ESXi.
Should (https://www.juniper.net/documentation/us/en/software/session-smart-router/docs/intro_downloading_iso/#downloading-an-iso) be updated?

docs/install_vmware_config.mdx
```diff
@@ -0,0 +1,149 @@
...
+5. Select the storage type and datastore to be used and then click **Next**. 
+
+	![Storage Type](/img/vmware_storage.jpg)
+
+#### Customize Settings
+
+6. Select **Virtual Hardware**.
+7. Expand the CPU selection, and configure the following settings:
+
+	- Set the number of CPU’s to at least 4.
```
MichaelBaj, line 67: You should reference the min specs here as well as reference our scaling guidelines.

docs/install_vmware_config.mdx
```diff
@@ -0,0 +1,149 @@
...
+	- Update the **Scheduling Affinity** to `0-3`.
+
+	Close the CPU list and configure the following under Virtual Hardware:
+
+	- Set memory to a minimum of 8GB.
+	- Set hard disk 1 to a minimum of 26GB.
```
MichaelBaj, line 73: 26GB is lower than our minimum guidelines.

docs/install_vmware_config.mdx
```diff
@@ -0,0 +1,149 @@
...
+	- Set the SCSI Controller 0 to `LSI Logic SAS`. 
+
+	![Virtual Hardware settings](/img/vmware_virt_hdwr.jpg)
+
+#### Configure Network Adapters
+
+8. Click the **Add network adapter** button two times, so you have a total of three network adapters available.
```
MichaelBaj, line 80: total of three? I would say, minimum of three. They should define however many they need.

docs/install_vmware_config.mdx
```diff
@@ -0,0 +1,149 @@
...
+	- Set hard disk 1 to a minimum of 26GB.
+	- Set the SCSI Controller 0 to `LSI Logic SAS`. 
+
+	![Virtual Hardware settings](/img/vmware_virt_hdwr.jpg)
+
+#### Configure Network Adapters
+
+8. Click the **Add network adapter** button two times, so you have a total of three network adapters available. 
+
+	- Expand each network adapter by clicking the small triangle and set the **Adapter Type** on each network adapter to be **E1000e**.
```
MichaelBaj, line 82: @jacobdionne I thought we don't recommend `e1000` driver any more. Even for ESXi.

MichaelBaj (changes requested):

docs/install_vmware_config.mdx
jacobdionne: I believe `vmxnet3` should be the best performant (or SR-IOV passthrough actually). However I think `e1000` is what was tested and qualified. There are some Jiras that suggest some issues were hit with vmxnet3. The ESXi version being used could also factor in what works and what doesn't

docs/install_vmware_config.mdx
Chr1st0ph3rTurn3r: Based on what you say above - ESXi is conductor managed - and qcow is only available for image-based installs, no the qcow cannot be used for ESXi.

MichaelBaj (approved):
Not having gone through this exercise myself, I can't speak the accuracy, but my comments have now been addressed.

Please note that there is a TODO in here and needs to be addressed prior to publishing.

PR #633, opened 2023-10-18
Service Area Troubleshooting Guide

MichaelBaj:


docs/ts_session_processing.md
```diff
@@ -0,0 +1,189 @@
...
+
+Connections per second deal with how quickly the SSR can create, modify and store a new session accepted by its configured policy. Every system has a maximum session processing rate that is governed by the platform on which it runs. In SSR software versions prior to 6.1, session processing is single-threaded and is rate-limited by CPU clock speed and overall CPU business (as the OS scheduler affords compute cycles to this task). In software versions 6.1 and beyond, session processing is multi-threaded and can take advantage of additional host cores.
+
+The session processing thread is responsible for session setup and modify operations. Modify operations are performed for a flow move (peer path failure or SLA violation) or with the use of application identification which updates the session after DPI. The session processing thread has a buffer queue for handling bursts of traffic or when the thread is “backed up”. As the queue fills up, latency increases for all packets in the queue. This will translate into end-to-end latency within the network. The system should be scaled with ample headroom to handle peak session processing load.
```
agrawalkaushik, line 19: > This will translate into end-to-end latency within the network.

This additional latency should only affect the first packet processing so maybe its worth noting that as well? Ideally for a TCP based session, this effect would just be felt on the initial session creation and once the session is underway the impact would disappear.

ryankoss: @ayungelson I thought that we continue to detour packets up to the service area until we're able to complete "first packet" processing allowing us to punch a hole through and create a session to the other side.  

This wouldn't be as a factor with TCP v UDP but then I don't know how deep into the weeds that we want to get with this...

docs/ts_session_processing.md
```diff
@@ -0,0 +1,189 @@
...
+
+Ideally the SSR should be operating at no more than 80% utilization of its session processing threads so that it can handle bursty conditions that would be triggered as a result of path migrations. When a path failures occur, or a service is no longer within SLA, the SSR needs to modify existing sessions to ensure they are traversing the optimal pathway. For systems carrying considerable load, this could be many thousands of sessions that need to be updated.
```
ryankoss, line 21: Remove Ideally

docs/ts_session_processing.md
```diff
@@ -0,0 +1,189 @@
...
+:::warning
+It is not recommended to set `session-processor-mode` to `manual` unless instructed by a member of Juniper support or engineering teams to ensure that this value is right-sized for the environment.
+:::
+
+```
+config authority router <router-name> node <node-name> session-setup-scaling   true
+config authority router <router-name> node <node-name> session-processor-mode  automatic
+```
+
+Once the configuration changes have been made and committed, a restart of the SSR is required for the changes to take effect.
```
ryankoss, line 36: Do want to say a restart of the SSR service?  Reason asking is that with us being a "device" now with hardware saying a restart of the SSR could have people think reboot of the SSR device.

docs/ts_session_processing.md
```diff
@@ -0,0 +1,189 @@
...
+ 7534 root      20   0  266.9g   8.8g 120940 S 55.0  3.5  34:42.19 SessionProc-02                                                       
+ 7535 root      20   0  266.9g   8.8g 120940 S 45.0  3.5  31:33.67 SessionProc-03                                                          
+ 7536 root      20   0  266.9g   8.8g 120940 S 35.0  3.5  27:27.56 SessionProc-04                                                          
+ 7537 root      20   0  266.9g   8.8g 120940 S 20.0  3.5  22:51.79 SessionProc-05
+```
+
+### Understanding CLI Output
+
+The "Service Area" is the colloquial name given to the business logic responsible for session processing. The set of stats related to `internal-application` of Service Area processing are valuable to understand system behavior. These stats are monotonically increasing scalar values, so one would need to compare deltas over two periods of time to get an accurate view of rate. In particular, `schedule-failure` and `sent-tmeout` are signs that the session processing is unable to keep up with load or unable to handle the entirety of bursty traffic patterns.
```
ryankoss, line 86: `sent-tmeout`

typo

docs/ts_session_processing.md
```diff
@@ -0,0 +1,189 @@
...
+Time series graphs of per-thread utilization are only available on SSR software starting with 6.1.
+:::
+
+The following graph will show time series thread utilization of session processing threads. This graph needs to be customized per router, per node, per process, and each SessionProc thread.
+When building this graph, make sure to include each of the running `SessionProc` threads.
+![Session Processing Thread Utilization](/img/ts_sp_session_processing_thread_utilization.png)
+`process/thread/cpu/usage`
+`Router <router-name> Node <node-name> Process Name 'highway' Thread Name 'SessionProc-XX'`
+
+As noted earlier, while monitoring session processing utilization is important it is not sufficient. Looking at the utilization of all other threads is important to understand how the CPU is scheduling work across all host processors. This is valuable to determine if other processes are starving out the session processing threads.
```
ryankoss, line 165: To ensure a wholistic view of session processing along with the stated above, looking that the utilization of all the other threads is important...

ryankoss (commented):
Also I'm not sure but do we want to discuss how packets/session get distributed to different session processing threads?  Could be an important distinction if people believe that this could help them in a hub-hub scaling case but ALL the traffic is coming from the same five-touple?

docs/ts_session_processing.md
ayungelson: @ryankoss yes, we do. I was debating that as well. Going off into the weeds... That's where the `service-area received existing-flows` comes in. This is for every "retransmitted" packet... It would be both, tcp syn retransmission if it took us forever to set up the session, or the flood of UDP packets. So it's hard to say whether it's system latency or not...

ryankoss: Have we come to consensus on what and how we'd like to document here

docs/ts_session_processing.md
```diff
@@ -0,0 +1,188 @@
+---
+title: Troubleshooting Session Processing
+sidebar_label: Session Processing
+---
+
```
ryankoss, line 5: Do we want to bring up the RSS hashing to Service-Area threads?  I can't find my older comment?

docs/ts_session_processing.md
agrawalkaushik: maybe my original comment only applies to TCP then, coz until we process the first SYN packet, the only other things that will punt up are more SYN retransmissions. But because of the nuance maybe not worth capturing. Thanks for entertaining my thought @ryankoss @ayungelson

ryankoss (approved):

docs/ts_session_processing.md
ryankoss: Honestly it's an interesting point... is this a note and or delineation that we want to cite?

PR #642, opened 2023-11-29
Static-NAT update

Chr1st0ph3rTurn3r:
Static-NAT update for non-svr traffic, per note from Chao

### DIFF
--- a/docs/config_nat.md
+++ b/docs/config_nat.md
@@ -106,6 +106,29 @@ exit
 
 ```
 
+### Non-SVR Traffic
+
+The `ingress-source-nat-pool` and the `source-nat` settings of `bidirectional-nat` are both `ingress source-nat` that are intended for SVR traffic targeted towards an inter-router peer. In this situation, the NAT will be applied on the ingress router rather than the final egress router.
+
+If you want to configure `ingress source-nat` for non-SVR traffic, for example LAN-to-LAN traffic traversing a single SSR router, you can use the configuration above. However, you must also disable `egress source-nat` at the service level by setting `service->source-nat` to `disabled` as shown below:
+
+```
+authority
+    service LAN-to-LAN
+        name LAN-to-LAN
+        description "LAN-to-LAN non-SVR traffic traversing a single SSR router"
+        source-nat disabled
+        scope private
+        security aes1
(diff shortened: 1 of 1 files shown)


docs/config_nat.md
```diff
@@ -106,6 +106,29 @@ exit
 
 ```
 
+### Non-SVR Traffic
+
+The `ingress-source-nat-pool` and the `source-nat` settings of `bidirectional-nat` are both `ingress source-nat` that are intended for SVR traffic targeted towards an inter-router peer. In this situation, the NAT will be applied on the ingress router rather than the final egress router.
```
MichaelBaj, line 111: I had to read the first sentence four times and I still don't think I grok it. This statement needs to be changed for readability.
Additionally, `ingress-source-nat-pool` is being referenced for the first time here. Definitions would help.

docs/config_nat.md
```diff
@@ -106,6 +106,29 @@ exit
...
+
+If you want to configure `ingress source-nat` for non-SVR traffic, for example LAN-to-LAN traffic traversing a single SSR router, you can use the configuration above. However, you must also disable `egress source-nat` at the service level by setting `service->source-nat` to `disabled` as shown below:
```
MichaelBaj, line 113: convention for referencing config path is to use a single right angle bracket between objects. e.g.
`service > source-nat`

MichaelBaj (changes requested):

docs/config_nat.md
chao-128t: @MichaelBaj below is the "annotation" for the first sentence.
Depending on when and where a packet gets source NAT'd, there are two types of source NAT - ingress and egress.

The `bidirectional-nat` config is in the form of `local-ip` and `remote-ip` pair.
- source-nat is performed for packet ingressing the interface where `bidirectional-nat` is configured.
- dest-nat is performed for packet egressing the interface where `bidirectional-nat` is configured.

In this sense, along with `ingress-source-nat-pool`, the `source-nat` part of `bidirectional-nat` falls into the ingress source-nat category and will be handled accordingly.

MichaelBaj (approved):

chao-128t (approved):

PR #645, opened 2023-12-06
Firewall Certification docs

Chr1st0ph3rTurn3r:


docs/sec_firewall_filtering.md
```diff
@@ -0,0 +1,268 @@
...
+2. Click Add.
+
+3. In the New Item window, name the rule and click SAVE.
+
+![New filter name](/img/bpf_image2.png)
+
+4. In the Filter Rule window, define:
+- The action - Deny discards any matching packets (what do they match? you haven't set any packet information in this filter). Permit allows packets that match the rule to bypass any additional rules, and passes the traffic.  
+- The Filter type - BPF is currently the only option. 
+- Berkeley Packet Filter - Identifies the filter to be applied. Validation confirms proper BPF syntax.
```
MichaelBaj, line 30: @ayungelson When we discussed this feature, you had mentioned that validation had not yet been added. Has that gap been closed?

docs/sec_firewall_filtering.md
```diff
@@ -0,0 +1,268 @@
...
+![Permit rule configuration](/img/bpf_image5.png) 
+
+7. After the Filter Rule list has been created, you can reorder the rules using the drop down menu to the left of the filter name. 
+
+![Reorder menu](/img/bpf_image6.png)
+
+This list determines the order in which filter rules are applied. 
+
+:::note
+The number of rules and the filtering operation is resource heavy and may impact performance.
```
MichaelBaj, line 47: Need @ayungelson to confirm, but I might word this statement differently:

The number and complexity of rules will have an impact on forwarding performance.

docs/sec_firewall_filtering.md
```diff
@@ -0,0 +1,268 @@
...
+exit
+```
+
+## ICMP 
+
+ICMP attributes have been updated for firewall protection. 
+
+#### Discard ICMP Echo Replies With No Request
+
+ICMP Echo Replies that arrive at the SSR are dropped if no corresponding request has been seen and the ICMP Aysnc Reply is set to drop. This is the new default behavior; no change in configuration is necessary.
```
MichaelBaj, line 92: New behavior as of when? This document should state that these features were introduced in 6.2.3-r2

docs/sec_firewall_filtering.md
```diff
@@ -0,0 +1,268 @@
...
+
+2. Scroll down to the IPv4 Option Filter button, and select it.
+
+![IPv4 Option Filter button](/img/ipv4_option_filter2.png)
+
+3. The IPv4 Options Filter window opens, indicating the `allow-all` action. Open the drop down and select `drop-all`.
+
+![IPv4 Option Filter window](/img/ipv4_option_filter3.png)
+
+If you do not want any allowed options, you do not have to proceed any farther. Return to the Authority level and continue with other configuration activities.
```
MichaelBaj, line 135: farter = further

docs/sec_firewall_filtering.md
```diff
@@ -0,0 +1,268 @@
...
+*admin@conductor.conductor# configure authority service-policy prefer-path-2 transport-
+state-enforcement strict
+*admin@conductor.conductor#
+```
+
+## TCP Half-Open Connection Limit
+
+Half-open TCP connections are those where the handshake has started but not completed. The SSR provides the ability to configure a limit to these half-open TCP connections. This provides protection against SYN flood DDoS attacks.
+
+The connection limit is configured at the router level (Authority > Router), and is unlimited by default. To set a limit, enter a numerical value in the `Half-Open Connection Limit` field in the Router Basic Information panel. When configured, the SSR tracks how many half-open sessions there are based on existing TCP session state and will deny any new TCP sessions once the limit has been reached.
```
MichaelBaj, line 199: @ayungelson It seems to me that we should add a WARNING admonition here regarding this feature.
Please correct as needed
::: warning
When the SSR reaches near its limit of half-open TCP connections, this may begin to prevent the establishment of "healthy" TCP traffic. Please ensure that this value is set appropriately for your network. Importantly, identify the devices that are creating half-open sessions.
:::

MichaelBaj (changes requested):

docs/sec_firewall_filtering.md
ayungelson: Yes, it has. We added config validation.

docs/sec_firewall_filtering.md
ayungelson: This statement isn't correct - this needs to be configured.
```
*admin@conductor.conductor# configure authority icmp-control icmp-async-reply drop
```

docs/sec_firewall_filtering.md
ayungelson: I don't think this setting should ever be used. 
I recommend removing this sentence entirely: `This provides protection against SYN flood DDoS attacks.`
And furthermore, I would just recommend not enabling it.

docs/sec_firewall_filtering.md
```diff
@@ -0,0 +1,268 @@
...
+        exit
+    exit
+exit
+
+*admin@conductor.conductor#
+```
+
+### Broadcast and Multicast Source Addresses
+
+Packets with broadcast or multicast source L2 addresses will now be dropped by default.
```
ayungelson, line 172: It's not just L2...
so maybe 
`Packets with broadcast or multicast source addresses` - or `Packets with broadcast or multicast source IP and MAC addresses`

docs/sec_firewall_filtering.md
```diff
@@ -3,7 +3,13 @@ title: Customizable Firewall Rules and Filters
 sidebar_label: Customizable Firewall Rules and Filters
 ---
 
-The following enhancements have been made to the SSR to provide more complete protection as a firewall. 
+As part of the security hardening and certification process, the SSR has implemented the following firewall features to provide a more secure platform for network traffic. Steps for configuration from both the GUI and the PCLI are provided.
```
MichaelBaj, line 6: I don't think we need justification for why we developed these features. To be a player in the Secure SD-WAN Market, these features are arguably table-stakes. Suggested rewording:

The SSR has firewall capabilities which offer a more secure platform for network traffic, blending a layer 3 and layer 4 firewall with SD-WAN. Steps for configuration from both the GUI and the PCLI are provided.

docs/sec_firewall_filtering.md
```diff
@@ -192,11 +200,21 @@ state-enforcement strict
...
+
+The connection limit is configured at the router level (Authority > Router), and is unlimited by default. To set a limit, enter a numerical value in the `Half-Open Connection Limit` field in the Router Basic Information panel. When configured, the SSR tracks how many half-open sessions there are based on existing TCP session state **and will deny any new TCP sessions once the limit has been reached**.
+
+:::warning
+When the SSR approaches the configured limit of half-open TCP connections, the establishment of **healthy** TCP sessions may be significantly impacted. Please ensure that this value is set appropriately for your network. More importantly, attempt to identify the devices that are creating half-open sessions.
+:::
+
+Additionally, if you require a limit for half-open TCP sessions, it may be helpful to consider the TCP session timeout value. The default timer is 1900000ms (roughly 31 minutes). If an application establishes a TCP socket that remains idle (without any keepalives) for longer than this value, the SSR will remove the session from the session table.
```
MichaelBaj, line 215: @ayungelson is this correct? 31 minutes for TCP initial state? That's got to be wrong. I thought TCP session timers were 30 min only upon establishment but initially very short during setup.

MichaelBaj (changes requested):

docs/sec_firewall_filtering.md
ayungelson: Half-open connections and TCP established timers really have nothing to do with each other... The initial timeout is defaulted to 10 seconds, although it can be changed, so maybe that's what this is referring to? And if the application actually establishes a TCP socket, it doesn't matter whether it's idle or not - it has nothing to do with the `half-open` connection limit. Those are fully opened and don't count against this limit.

MichaelBaj: Please correct me if I am wrong, but TCP initial timers have a direct impact on half-open connection limit.
If the initial connection timer is 10 minutes and the number of half-open connections allowed is 1000 (presuming all half-open connections are from a bad actor), then the system will not be able to handle a new TCP connection for 10 minutes.
If the initial connection timer is 10 seconds and the number of half-open connections allows is 1000 (presuming all half-open connections are from a bad actor), then the system will be able to process the next TCP session in 10 seonds.
Therefore, it is important to tune both parameters: Queue depth and Queue wait time before discard.

ayungelson: @MichaelBaj yes - initial time matters of course. That's typically not tuned, and is defaulted to 10 seconds. If we wanted to mention this, then we should change the wording, something like this:
```
Additionally, if you require a limit for half-open TCP sessions, it may be helpful to consider the initial TCP session timeout value. The default timer is 10 seconds. If an application fails to establish a TCP socket, the sessions that are in that state will still remain on the SSR for that initial timeout. 
```

PR #648, opened 2023-12-15
Adding details regarding vector costs

MichaelBaj:


docs/concepts_glossary.md
```diff
@@ -489,6 +489,78 @@ network-interface  mpls1
 exit
 ```
 
+When defining vectors within a service-policy for association within a neighborhood, network-interface or service-route, each will be set with a default priority of `ordered`. The order in which they are configured define the precedence for which a path should be used. The use of ordered vectors triggers the conductor to automatically generate a value for the vector cost. This numbering starts at one and increases by an order of magnitude (1, 10, 100, 1000, 10000, 100000, 999999). Ordered lists longer than six vectors will result in additional vectors autogenerated with the highest configurable cost of 999999 which would result in these vectors being considered equal cost. Therefore, for service-policies with more than six vectors it is required to manually set the cost.
+
+It is possible to mix explicit vector values with implicit, ordered values, although this is not recommended. For cases where this type of configuration is necessary, the follow example illustrates the system's behavior when doing so:
```
laneshields, line 494: Do we even want to open this door? I guess where would this be necessary?

MichaelBaj: I don't necessarily want to, but at the same time, we rightly got called out for not providing details for how order cost is defined. The same holds true for interspersing explicit and implicit cost. IMO, we are obligated to describe the system behavior.
Arguably all of this could have been avoided if we could just print out the generated values right from the config.

docs/concepts_glossary.md
```diff
@@ -489,6 +489,78 @@ network-interface  mpls1
 exit
 ```
 
+When defining vectors within a service-policy for association within a neighborhood, network-interface or service-route, each will be set with a default priority of `ordered`. The order in which they are configured define the precedence for which a path should be used. The use of ordered vectors triggers the conductor to automatically generate a value for the vector cost. This numbering starts at one and increases by an order of magnitude (1, 10, 100, 1000, 10000, 100000, 999999). Ordered lists longer than six vectors will result in additional vectors autogenerated with the highest configurable cost of 999999 which would result in these vectors being considered equal cost. Therefore, for service-policies with more than six vectors it is required to manually set the cost.
```
Chr1st0ph3rTurn3r, line 492:
```suggestion
When defining vectors within a service-policy for association within a neighborhood, network-interface, or service-route, each will be set with a default priority of `ordered`. The order in which they are configured define the precedence for which path should be used. The use of ordered vectors triggers the conductor to automatically generate a value for the vector cost. This numbering starts at one and increases by an order of magnitude (1, 10, 100, 1000, 10000, 100000, 999999). Ordered lists longer than six vectors will result in additional vectors autogenerated with the highest configurable cost of 999999 which would result in these vectors being considered equal cost. Therefore, for service-policies with more than six vectors it is required to manually set the cost.
```

Chr1st0ph3rTurn3r (changes requested):

docs/concepts_glossary.md
MichaelBaj: As a compromise, we can add this content in a collapsible section so that it does not draw attention.
Attached are rendered examples of the implementation with the collapsed section closed and open:
Screenshot 2023-12-15 at 2 01 29 PM
Screenshot 2023-12-15 at 2 01 44 PM

Chr1st0ph3rTurn3r (approved):
I like the collapsable section. I haven't seen that in years.

avinashbhat80 (approved):

PR #653, opened 2023-12-29
Single Router Migration per DOCS-284

Chr1st0ph3rTurn3r:


docs/howto_router_migration.md
```diff
@@ -0,0 +1,38 @@
...
+title: Router Migration
+sidebar_label: Router Migration
+---
+
+This process is used to migrate an existing conductor-managed router to a new conductor and infrastructure. From the PCLI, the command can be run from the existing conductor to automatically migrate the router to the new conductor, or it can be run from the router itself. Before perfoming the procedure below, the new conductor must be installed and configured, and both the old and the new conductors must be reachable. For steps to install a conductor, see [Single Conductor Interactive Installation](single_conductor_install.mdx).
+
+### PCLI
+
+1. Export the router configuration from the old conductor, and import and commit the router configurations to the new conductor. Alternatively, you may be able to simply copy and paste the configuration to the new conductor. 
+2. Run the `migrate` command as follows:
```
ian-h-chamberlain, line 11: It might be good to clarify that only one of the two options is necessary. This reads to me as if you would need to do both in order.

ian-h-chamberlain (approved):

docs/howto_router_migration.md
```diff
@@ -7,28 +7,30 @@ This process is used to migrate an existing conductor-managed router to a new co
...
 
-1. Export the router configuration from the old conductor, and import and commit the router configurations to the new conductor. Alternatively, you may be able to simply copy and paste the configuration to the new conductor. 
-2. Run the `migrate` command as follows: 
+The `migrate` command automates the process of migrating a router to a new conductor. 
+
+Run the `migrate` command as follows: 
  - From the old conductor: `migrate conductor <new-address> router <router>`
  - From the router: `migrate conductor <new_address>`
 
+Alternatively, you can export the router configuration from the old conductor, and import and commit the router configurations to the new conductor. Or depending on the complexity of the configuration, you may be able to simply copy and paste the router configuration to the new conductor.
```
ian-h-chamberlain, line 16: Hmm, this isn't quite what I had in mind. The configuration export/import is always necessary, as far as I know — but either of the two bulleted `migrate conductor` commands mentioned could be used. I'd probably write something like this:

```suggestion
1. Export the router configuration from the old conductor, and import and commit the router configurations to the new conductor. Depending on the complexity of the configuration, you may be able to simply copy and paste the router configuration to the new conductor.
2. Run either of the following `migrate` commands:
 - From the old conductor: `migrate conductor <new-address> router <router>`
 - From the router: `migrate conductor <new_address>`
```

ian-h-chamberlain (dismissed):

docs/howto_router_migration.md
```diff
@@ -0,0 +1,39 @@
+---
+title: Router Migration
+sidebar_label: Router Migration
+---
+
+This process is used to migrate an existing conductor-managed router to a new conductor and infrastructure. From the PCLI, the command can be run from the existing conductor to automatically migrate the router to the new conductor, or it can be run from the router itself. Before perfoming the procedure below, the new conductor must be installed and configured, and both the old and the new conductors must be reachable. For steps to install a conductor, see [Single Conductor Interactive Installation](single_conductor_install.mdx).
```
MichaelBaj, line 6: "... the command can be run from the existing conductor to automatically migrate the router to the new conductor, or it can be run from the router itself"
Why the two different approaches? Why is one better than the other? Under what circumstances would you use one vs. the other? Can this be spelled out? If there is no difference, best to document a procedure using one approach only. Then add a footnote that the process can be done the other way, as needed.

docs/howto_router_migration.md
```diff
@@ -0,0 +1,39 @@
...
+
+### PCLI
+
+The `migrate` command automates the process of migrating a router to a new conductor. 
+
+1. Export the router configuration from the old conductor, and import and commit the router configurations to the new conductor. Depending on the complexity of the configuration, you may be able to simply copy and paste the router configuration to the new conductor.
```
MichaelBaj, line 12: I don't believe there is a way to export only a specific router's configuration @ian-h-chamberlain.
Can we enumerate the exact steps needed to obtain only the router's specific config?
Also, what about references to config outside of the router config model? Presumably those need to be copied over as well, else validation will fail.

MichaelBaj (changes requested):

docs/howto_router_migration.md
ian-h-chamberlain: Functionally, both methods should be the same in terms of what actually happens to the router. I'd guess that driving it from the conductor is normally preferred for convenience, but if the user wanted to debug / monitor the system during migration they might want to drive it from the router (since it will at some point become inaccessible from the old conductor as part of the migration).

docs/howto_router_migration.md
ian-h-chamberlain: That's true, and what it actually means to obtain a specific router's config probably varies per-deployment, since as you mention datamodel references to peers etc. would depend on the existing configuration. I'm not sure there's really a one-size-fits-all procedure for copying the configuration, but two options but the first thing that comes to mind would be something like this:

1. From the router PCLI (so it only includes config that applies to that router), `show config running flat > /tmp/router_config.txt`
2. Remove `conductor-address` from the config so that "import" doesn't overwrite it, and adjust any routing configuration that the router will need to be able to reach the new conductor
3. Copy `/tmp/router_config.txt` to the new conductor
4. In new conductor PCLI: `< /tmp/router_config.txt`
5. Perform migration with `migrate` command.

Step 2 is the most nebulous since it will depend on how the router reaches the new Conductor — if no special routing configuration is required then it might be trivial, but if management-over-forwarding is used or anything like that it could be more involved and deployment-specific.

Chr1st0ph3rTurn3r: I just did something similar to test my own theory about exporting just the router config. I opened the pcli for a router, did a show config running, and then ran 
```
admin@128t-west.128t-west# export config running west-router
✔ Exporting configuration...
Successfully exported configuration: /etc/128technology/config-exports/west-router.gz
```

ian-h-chamberlain (approved):

MichaelBaj (approved):

PR #719, opened 2024-05-21
Adding clarification to since command for rate stats

MichaelBaj:
DOCS-405 #time 30min

### DIFF
--- a/docs/concepts_metrics.md
+++ b/docs/concepts_metrics.md
@@ -45,6 +45,8 @@ IPFIX Stats
 Completed in 0.06 seconds
 ```
 
+The since argument is a general utility which performs a delta between the two values supplied by the arguments. The primary use is for cumulative metrics - ones that keep incrementing. Some stats are stored as a rate, which are a per 5 second view of the stat. A rate valeu in the past may be higher then the current rate (or rate at the selected time) and therefor a negative value may be returned.
+
 ## Session Establishment Metrics
 A key indicator of application performance is the time it takes to establish the TCP session between client and server. This is effectively the time it takes to get to the first data packet between endpoints. This metric is more telling than packet transmission rates because it is directional and end to end. Importantly, this information can be used as a measure of SLA to influence path selection.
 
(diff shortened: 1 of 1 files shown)


docs/concepts_metrics.md
```diff
@@ -45,6 +45,8 @@ IPFIX Stats
 Completed in 0.06 seconds
 ```
 
+The since argument is a general utility which performs a delta between the two values supplied by the arguments. The primary use is for cumulative metrics - ones that keep incrementing. Some stats are stored as a rate, which are a per 5 second view of the stat. A rate valeu in the past may be higher then the current rate (or rate at the selected time) and therefor a negative value may be returned.
```
Chr1st0ph3rTurn3r, line 48:
```suggestion
The `since` argument is a general utility which performs a delta between the two values supplied by the arguments. The primary use is for cumulative metrics - ones that keep incrementing. Some stats are stored as a rate, which is a per 5 second view of the stat. A rate value in the past may be higher than the current rate (or rate at the selected time) and therefore a negative value may be returned.
```

Chr1st0ph3rTurn3r (changes requested):

Chr1st0ph3rTurn3r (approved):

docs/concepts_metrics.md
```diff
@@ -45,6 +45,8 @@ IPFIX Stats
 Completed in 0.06 seconds
 ```
 
+The `since` argument is a general utility which performs a delta between the two values supplied by the arguments. The primary use is for cumulative metrics - ones that keep incrementing. Some stats are stored as a rate, which is a per 5 second view of the stat. A rate value in the past may be higher than the current rate (or rate at the selected time) and therefore a negative value may be returned.
```
gregschrock, line 48: The description is accurate, but I'm confused by the location.
1. I would not expect the since command to show up here as it is least relevant for the in memory metrics. They do not have history, so they should only really have a caveat saying they'll only show the current value when used with `since`.
2. Rate metrics and in memory metrics are not related, so the section header makes this out of place.

It seems like we should have a section just for the `show since` command where we can put these caveats.

MichaelBaj: Appreciate the input. Put up PR https://github.com/128technology/docs/pull/721 to provide this correct. Additional input is requested.

PR #730, opened 2024-06-12
Per commitment to customer, 6.2. release is to be extended to LTS

MichaelBaj:


MichaelBaj:
This was committed solely for one customer. Would like to honor this commitment for the one customer without supporting for all others.

PR #743, opened 2024-07-12
removing one BYOL bullet point that was mistakenly added

Chr1st0ph3rTurn3r:
and one that was never removed.


MichaelBaj (approved):
This request to remove must have come from @agrawalkaushik. If so, I approve.

PR #757, opened 2024-09-16
Add BYOL release notes

agrawalkaushik:


docs/intro_installation_byol_azure_mist.md
```diff
@@ -370,3 +370,33 @@ New-AzResourceGroupDeployment -ResourceGroupName <your-resource-group-name> `
 :::important
 When logging to the Linux instance via SSH use the default username of `t128` and the private key specified in the **Admin Public Key Data** field.
 :::
+
+## Release Notes
```
MichaelBaj, line 374: You are adopting the plugin release notes model wherein the guide has a trailing release notes section.
What does 1.1.0 refer to in this regard? Is this the BYOL image? How does a customer understand what version they are running?
IMO, we should have a separate release notes category for BYOL and list this information there instead. The configuration guides should be separate from the release notes.
@Chr1st0ph3rTurn3r This is consistent with our migration to the Juniper docs format, correct?

docs/intro_installation_byol_azure_mist.md
```diff
@@ -370,3 +370,33 @@ New-AzResourceGroupDeployment -ResourceGroupName <your-resource-group-name> `
...
+
+- **I95-58274** Azure instances can sometimes fail to successfully onboard on first boot.
+
+
+### Release 1.0.0
+
+**Release Date:** Sep 2nd, 2024
+
+#### New Features and Improvements
+- **I95-48255** BYOL Support for Mist Managed router
```
Chr1st0ph3rTurn3r, line 400: Put everything in one entry:
- **I95-48255 BYOL Support for Mist Managed router:** Created a new BYOL offering in Azure, to provide seamless onboarding for MIST managed Session Smart Routers. The offer enables various modes of operation using ARM templates, image-based, cloud-init based, and manual onboarding modes to best fit the end user workfow.

docs/intro_installation_quickstart_byol_mist_aws.md
```diff
@@ -326,3 +326,29 @@ If the EC2 instance deployed for the Session Smart software does not have access
...
+- **I95-58274** Serial console does not show any output in AWS BYOL image
+
+  _**Resolution:**_ The boot line for the BYOL image now contain the appopriate options ot enable the console to operate correctly.
+
+### Release 1.0.0
+
+**Release Date:** Sep 1st, 2024
+
+#### New Features and Improvements
+- **I95-48255** BYOL Support for Mist Managed router
```
Chr1st0ph3rTurn3r, line 352: Put everything in one entry:

- **I95-48255 BYOL Support for Mist Managed router:** Created a new BYOL offering in AWS, to provide seamless onboarding for MIST managed Session Smart Routers. The offer enables various modes of operation using CloudFormation templates, AMI-based, cloud-init based, and manual onboarding modes to best fit the end user workfow.

Chr1st0ph3rTurn3r (changes requested):

docs/intro_installation_byol_azure_mist.md
Chr1st0ph3rTurn3r: It is, and I thought about suggesting a release notes section for AWS and Azure, since they have separate "guides" in the docs now, and will be getting more with the 6.3 release (BYOL for conductor). Then I thought if there was a need to get it out right away, I would let it go as is for now and reformat the docs after this goes out as part of the 6.3 release. If Kaushik has the cycles and wants to do it now, that is fine with me. If not, I will get it reformatted over the next week or so.

agrawalkaushik: > What does 1.1.0 refer to in this regard? Is this the BYOL image? How does a customer understand what version they are running?

Yes. The version is visible on AWS. For Azure, its what they would by default pick the latest but for some reason if they needed/wanted to stick to an older version, they would get them via API and use in their templates.

agrawalkaushik: I'll update it to add a new sidebar for BYOL.

PR #776, opened 2024-10-10
Added clarity around the use and need of max-way-points

MichaelBaj:


avinashbhat80 (approved):

docs/concepts_waypoint_ports.md
```diff
@@ -27,10 +27,12 @@ Thus, each wayport on an SSR can initiate 24,574 sessions to each of 24,574 uniq
...
+
+It is important to note that despite the configured [port-range](config_reference_guide.md#port-range), the SSR will at most use 50K of those ports by default. The configuration [max-way-points](config_reference_guide.md#max-way-points) allows the SSR software to scale respective to the platform and the administrator's needs. A `max-way-points` setting that is smaller than the number of sessions traversing a peer path will result in port exhaustion. A `max-way-points` setting that is too large will result in wasted system resources. Best practice is to tune both the `port-range` and `max-way-points` to a factor of 1.5x of the max number of sessions traversing a peer path.
```
Chr1st0ph3rTurn3r, line 38:
```suggestion
It is important to note that despite the configured [port-range](config_reference_guide.md#port-range), the SSR will at most use 50,000 of those ports by default. The configuration [max-way-points](config_reference_guide.md#max-way-points) allows the SSR software to scale respective to the platform and the administrator's needs. A `max-way-points` setting that is smaller than the number of sessions traversing a peer path will result in port exhaustion. A `max-way-points` setting that is too large will result in wasted system resources. Best practice is to tune both the `port-range` and `max-way-points` to a factor of 1.5x of the max number of sessions traversing a peer path.
```

Chr1st0ph3rTurn3r (changes requested):

docs/concepts_waypoint_ports.md
```diff
@@ -27,10 +27,12 @@ Thus, each wayport on an SSR can initiate 24,574 sessions to each of 24,574 uniq
...
\ No newline at end of file
+Limiting this range will make your firewall team happy without putting too much of a crimp in your ability to SVR sessions. In fact, we have some customers that have limited their waypoint port allocation to only 10 ports on their head end routers! At face value this seems like it would severely restrict the number of unique, concurrent sessions that could be established by a peer, but in practice it doesn't. If a branch keeps its range to the default value of ports (49,148 ports), which can be sent to each of ten ports, you still wind up with a theoretical maximum of over 120,000 concurrent sessions *in each direction* to that branch over that adjacency.
```
chao-128t, line 36: - The default **local** port range is 16384 ~ 65534, 49151 ports in total, not 49148.

> the default value of ports (49,148 ports)

- The default **remote** port range is 16385 ~ 65533, 49149 ports in total.

- When (remote) port-range is not configured, local ports are even, remote ports are odd, the total combination is 24576 (even local ports) * 24575 (odd remote ports) = 603,955,200
- When (remote) port-range is configured, local ports are continuous within that 16384 ~ 65534 range, remote ports are also continuous within the configured range, so the total combination is 49151 * `num of ports within the range`. In the doc's example, the total combination is 49151 * 10 = 491,510.

MichaelBaj: What do you mean by "local" vs "remote" port ranges? Are the settings per neighborhood, and therefore shared by both nodes? Or is that not the case and for any given neighborhood, there is both a local and remote, because the setting applies to that device specifically?

chao-128t: For a given SVR session, waypoint ports are allocated by the ingress router. So, the notion of local vs. remote is from the ingress router's perspective. The local port does not have a configurable range only the remote port does. It tells the ingress router what ports are opened on the egress router for SVR sessions.

The port range is per neighborhood per network-interface and only shared across an HA pair if the interface is redundant.

The `neighborhood->port-range` on the hub is propagated to `adjacency->port-range` on the branch through adjacency generation by the conductor.

Chr1st0ph3rTurn3r (approved):

chao-128t (approved):

PR #847, opened 2025-06-03
Add discussion of NAT behavior for EoSVR

laneshields:
Added discussion of NAT behavior

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -52,6 +52,7 @@ module.exports = {
         "type": "category",
         "label": "WAN Assurance Plugin",
         "items": [
+          "release_notes_wan_assurance_plugin_3.11",
           "release_notes_wan_assurance_plugin_3.10",
           "release_notes_wan_assurance_plugin_3.9",
           "release_notes_wan_assurance_plugin_3.8",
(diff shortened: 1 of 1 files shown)


docs/concepts_EthOverSVR.md
```diff
@@ -69,6 +69,9 @@ Service Vxlan_To_R1
 ```
 If more addresses are involved, they are added to the existing service. All VxLAN traffic is sent over SVR with new fields in ebedded in the metadata. 
 
+### Ethernet Over Secure Vector Routing across a NAT boundary
+
+Due to differences in the nature of EoSVR vs regular SVR, standard methods for stuck session detection and prevention may not function as well for EoSVR traffic. In scenarios where a spoke router needs to sit behind a NAT from the perspective of the hub it is recommended to have the NAT device set as a 1:1 NAT to the SSR as opposed to doing an NAPT.
```
MichaelBaj, line 74: Recommend changing "stuck session detection and prevention" to: [Session Recovery Detection](https://www.juniper.net/documentation/us/en/software/session-smart-router/docs/config_session_recovery/#session-recovery-detection).

MichaelBaj (changes requested):

docs/concepts_EthOverSVR.md
Chr1st0ph3rTurn3r:
```suggestion
Due to differences in the nature of EoSVR versus regular SVR, standard methods for [Session Recovery Detection](config_session_recovery.md#session-recovery-detection) and prevention may not function as well for EoSVR traffic. In scenarios where a spoke router needs to sit behind a NAT in relation to the hub, it is recommended to set the NAT device to 1:1 NAT to the SSR, rather than implementing NAPT.
```

PR #956, opened 2026-04-07
Provide more details and examples for certification creation

MichaelBaj:


Chr1st0ph3rTurn3r (dismissed):

.github/AGENTS.md
```diff
@@ -0,0 +1,31 @@
...
+---
+## ⚡ Agent Routing
+
+**Any agent that receives a request** must apply this routing check as its **first action**, before any analysis, file read, or tool call:
+
+**How to invoke a sub-agent:** Use the `runSubagent` tool with the agent name and pass the original user message verbatim as the argument. Do not summarise or reframe it. Sub-agent invocation is always preferred over asking the user to switch; manual switch is a fallback only when `runSubagent` is unavailable.
+
+**When intent is ambiguous:** Use `vscode/askQuestions` to clarify before routing. A 30-second clarification saves minutes of backtracking. Do not assume scope, bundle path, or symptom — ask first.
+
+> **Why this matters:** Accepting an investigation task in the wrong agent bypasses the investigation loop, playbook routing, output contract, and reporting standard — producing inline chat findings with no `analysis.md`, no `debug.md`, and no evidence ledger.
```
agrawalkaushik: likely a copy/paste error but this wont apply in this repo.

.github/AGENTS.md
```diff
@@ -0,0 +1,31 @@
...
+
+---
+## Artifact Types
```
agrawalkaushik: same here - this section will confuse the agent since the referenced artifacts don't exist in this repo

docs/config-custom-certs.md
```diff
@@ -284,4 +327,26 @@ Once the certificate is successfully ingested, verify that the certificate was a
 3. Verify that `/etc/128technology/pki/custom_ssr_peering.pem` exists on disk.
   `ls -l /etc/128technology/pki/custom_ssr_peering.pem`
 
-Enhanced Security Key Management can now use the private key and certificate.
+### Activate the Certificate in Configuration
```
laneshields-jnpr, line 325: We kind of jump to this without discussing how to get the signed cert on the router. I was able to POST that via:
  POST /api/v1/router/{router_name}/node/{node_name}/certificate
  {"name":"custom_ssr_peering","certificate":"-----BEGIN CERTIFICATE-----..."}

But I am running 7.1.4 and I see notes above about "remove when 7.0.4 is releaed" and I am not sure if that is associated.

docs/config-custom-certs.md
```diff
@@ -55,6 +55,16 @@ The following are some details of certificate security.
 It is necessary for all of the REST APIs to use the name `custom_ssr_peering` in order for this private key and certificate to be visible and usable by Enhanced Security Key Management in 7.0. This is a reserved name specifically used by the Enhanced Security Key Management feature.
 :::
 
+:::tip Swagger API Reference
+Before issuing API calls, review the full parameter options and per-router/per-node endpoint variants in the Swagger documentation available at:
+
+`https://<conductor-ip>/api/v1/swagger`
+
+Search for `/api/v1/private-key`, `/api/v1/certificate-request`, and `/api/v1/certificate` to see all supported parameters, algorithm options, key sizes, and the router-scoped and node-scoped path variations.
```
laneshields-jnpr: FWIW, I don't see the router/node specific paths in our swagger page, just the three listed here. I needed Eng to point out to me on slack the paths for the router/node specific ones.

docs/config-custom-certs.md
```diff
@@ -165,13 +175,18 @@ Store the value of the token in a file called `token.txt` for use later.
...
 ### Issue a Private-key Creation Request
 
-The goal of this workflow is to ensure that the private key of the SSR never leaves the SSR. To do so, we need to instruct the SSR to create a private key. To accomplish this, we provide the SSR some details, including:
+The goal of this workflow is to ensure that the private key of the SSR never leaves the SSR. To do so, instruct the SSR to create a private key by providing the algorithm, key size, and key name.
+
+:::important Per-Router / Per-Node Execution
+This API call must be made **separately for every router** participating in ESKM and **separately for every node** in an HA pair. Use the endpoint appropriate to your target:
+
+- **Router-scoped:** `POST https://<conductor-ip>/api/v1/router/<router-name>/private-key`
+- **Node-scoped (HA):** `POST https://<conductor-ip>/api/v1/router/<router-name>/node/<node-name>/private-key`
```
laneshields-jnpr, line 183: I used the node-scoped one in my test even though it wasn't HA. Do we want to just recommend that all around to avoid confusion? Actually, a quick test shows I get a 404 when trying the router-scoped one.

docs/config-custom-certs.md
laneshields-jnpr, line 265: I might use the full /router/{router_name}/node/{node_name} path in the example or point out that the API below needs to be done directly to the router's API and not through the conductor.

docs/enhanced-sec-key-mgmt.md
```diff
@@ -63,6 +63,60 @@ To provide thorough, end-to-end security, the use of a trusted and provisioned c
 The user provided certificates and signing authority must be in place before configuring Enhanced Security Key Management. If they are NOT in place prior to configuration and are added afterwards, then the SSR service must be restarted in order to pick up the changes.
 :::
 
+### Certificate Provisioning Workflow
+
+When using CA-signed certificates, the end-to-end provisioning sequence must be completed for **every router** that participates in ESKM and **for every node** in any HA pair:
+
+1. Authenticate to the Conductor REST API (`POST /api/v1/login`) and store the bearer token.
+2. Generate a private key on the target router or node (`POST /api/v1/router/<router>/private-key` or `POST /api/v1/router/<router>/node/<node>/private-key`) — the key never leaves the SSR.
```
laneshields-jnpr: only the second API example shown here seems valid

docs/enhanced-sec-key-mgmt.md
```diff
@@ -63,6 +63,60 @@ To provide thorough, end-to-end security, the use of a trusted and provisioned c
...
+3. Generate a CSR on the same target (`GET /api/v1/router/<router>/certificate-request` or with `/node/<node>/`) using a `common_name` that matches that router's `peering-common-name`.
```
laneshields-jnpr: only the second API example shown here seems valid

docs/enhanced-sec-key-mgmt.md
```diff
@@ -63,6 +63,60 @@ To provide thorough, end-to-end security, the use of a trusted and provisioned c
...
+4. Submit the CSR to your Certificate Authority and obtain a signed certificate.
+5. Ingest the signed certificate back into the SSR (`POST /api/v1/router/<router>/certificate` or with `/node/<node>/`).
```
laneshields-jnpr: only the second API example shown here seems valid

laneshields-jnpr (changes requested):

docs/config-custom-certs.md
madamsJuniper: Yes, I believe we need to remove the router scoped call here.

docs/config-custom-certs.md
```diff
@@ -183,16 +198,32 @@ Create the following file (updated to the customers algorithm/key size preferenc
...
 
-Now that we have the key request details ironed out, we can issue the request:
+**Router-scoped example** (single-node router or non-HA):
+
+```
+curl -k -X POST https://10.27.35.89/api/v1/router/combo-east/private-key \
+  -H "Content-Type: application/json" \
+  -H "Authorization: Bearer $(cat token.txt)" \
+  -d @key_request.json
+```
```
madamsJuniper: This example of a router-scoped api call is incorrect and needs to be removed

docs/config-custom-certs.md
```diff
@@ -205,14 +236,26 @@ This requirement is lifted starting with SSR 7.1.0.
 :::
 <!---remove this note when 7.0.4 is released and the restriction is lifted for 7.0.x --->
```
madamsJuniper, line 227: As this says, we should remove this now that 7.1 has shipped and we fully support custom certs

docs/enhanced-sec-key-mgmt.md
madamsJuniper: correct, the `router` level API does not exist and needs to be removed.

docs/enhanced-sec-key-mgmt.md
madamsJuniper: as seen throughout, the router level API does not exist and needs to be removed

docs/enhanced-sec-key-mgmt.md
madamsJuniper: same as above

docs/enhanced-sec-key-mgmt.md
```diff
@@ -165,7 +228,11 @@ config
     authority
         enhanced-security-key-management  true
 ```
-2. Configure a `peering-common-name` on each router. In a secure environment, the router name should never be sent between routers as plaintext in BFD messages. The `peering-common-name` is an alias that identifies the router and is configured at the router level. When `enhanced-security-key-management` is configured, it is validated against the peering-common-name from the ceritficate, and integrated into the auto-generated adjacencies list for the peers of the router from the neighborhood configuration.
+2. Configure a unique `peering-common-name` on each router. In a secure environment, the router name should never be sent between routers as plaintext in BFD messages. The `peering-common-name` is a **unique alias per router** that identifies the router and is configured at the router level. When `enhanced-security-key-management` is configured, it is validated against the `peering-common-name` from the certificate, and integrated into the auto-generated adjacencies list for the peers of the router from the neighborhood configuration.
```
madamsJuniper, line 233: This is good for now, but we will need to add a section on how to set the custom SAN type for the SSR once that bit of code goes in.

The uniqueness of each node's cert must currently come from one of the other fields in the cert (O/OU, etc.)

docs/enhanced-sec-key-mgmt.md
```diff
@@ -63,6 +63,60 @@ To provide thorough, end-to-end security, the use of a trusted and provisioned c
...
+**Example mapping for a two-router deployment:**
+
+| Router | `peering-common-name` in config | `common_name` in CSR | `name` in all API requests |
+|---|---|---|---|
+| combo-east | `east-alias` | `east-alias` | `custom_ssr_peering` |
+| combo-west | `west-alias` | `west-alias` | `custom_ssr_peering` |
+
+#### Activating the Certificate in Configuration
+
+After the certificate file is present on disk at `/etc/128technology/pki/custom_ssr_peering.pem`, you must register it in the SSR authority configuration. **The certificate will not be loaded by Enhanced Security Key Management until this step is complete, even if the `.pem` file exists on disk.**
```
madamsJuniper: Another caveat that this is 7.0 specific.  In 7.1 we allow any name, not just custom_ssr_peering.  As such, the private-key, CSR, and cert will be written to disk using the `name` from the API, but with the appropriate file type (`.key`, `.csr`, and `.pem` respectively).

docs/config-custom-certs.md
madamsJuniper: Good point - that is the correct API for ingesting the cert, and we should add it here.

madamsJuniper (changes requested):

.github/AGENTS.md
MichaelBaj: Should not have included this file in this PR. Removed.

.github/AGENTS.md
MichaelBaj: Should not have included this file in this PR. Removed.

docs/config-custom-certs.md
MichaelBaj: IIUC there only exists these 3 APIs and they are node-specific.

MichaelBaj:
I believe I have addressed all of the review feedback. Please re-review for accuracy.

docs/config-custom-certs.md
```diff
@@ -231,8 +264,9 @@ This example represents the minimum requirements. Any of the following additiona
 
 2. Issue the CSR request to the SSR:
 
+The API needs to be executed directly against the router's API and not through the conductor.
```
madamsJuniper, line 267:
```suggestion
```
This statement is incorrect and can be removed

madamsJuniper (changes requested):
Looks good aside from one requested change

docs/config-custom-certs.md
MichaelBaj: I put this statement in the document based on Lane's comment: https://github.com/128technology/docs/pull/956#discussion_r3047940634

Is his statement incorrect, or did I misunderstand?

laneshields-jnpr (approved):

docs/config-custom-certs.md
laneshields-jnpr: I can see why it was confusing. The example previously had /api/v1/certificate-request as the URL. If this is the URL it would need to be executed directly on the router's API and not through the Conductor. If through the Conductor you need the full context URL of /api/v1/router/{router_name}/node/{node_name}. I did not state what I meant clearly, though. I think the recommendation and typical usage should be to go through Conductor and use the full context API.

madamsJuniper: Ah, my point was that `/api/v1/router/{router_name}/node/{node_name}` can be issued to either the conductor or the node in question.  If it was targeted directly to the node, the router name/node name are sanity checked and allowed to be proxied locally.  I agree with Lane's comment though, the typical recommendation should be to go through the conductor and use the full context API.

MichaelBaj: Added the admonition:

:::note
The API request `/api/v1/router/{router_name}/node/{node_name}` can be issued to either the conductor or the node. If it was targeted directly to the node, the router name/node name are validated by the device and allowed to be proxied locally. The best practice is for all API requests to be routed through the conductor and use the full context API.
:::

madamsJuniper (approved):

PR #958, opened 2026-04-10
Swap broken-link-checker for linkinator

MichaelBaj:


docs/install_qcow2_deployment.md
```diff
@@ -9,14 +9,12 @@ For successful SSR qcow2 deployment, a cloud-init (or nocloud) configuration **m
 
 Download the 128T package-based qcow2 image using the link below:
 
-<!-- markdown-link-check-disable-next-line -->
```
Chr1st0ph3rTurn3r, line 12: What happens when it tries to access a good link, but does not have the credentials to access the directory? Will it still report a broken link?

Chr1st0ph3rTurn3r (approved):

docs/install_qcow2_deployment.md
MichaelBaj: It depends on how the page is implemented. I can be either with a 403 (failure) or a redirect to a page asking for credentials. If you look at the linkinator.config.json file, you will see that I added an exclusion for our repos. I did this to be consistent with our prior approach.

PR #993, opened 2026-04-30
initial gcp docs

haberkornsam:
Initial GCP docs

TODO:

- [x] Add screenshots of deployment page


docs/intro_installation_byol_gcp_conductor.md
```diff
@@ -0,0 +1,409 @@
...
+
+For the latest information about SSR BYOL offerings, refer to the [Cloud Images BYOL Release Notes](release_notes_byol.md).
+
+After selecting the BYOL plan, continue to [Session Smart Conductor Deployment](#session-smart-conductor-deployment) to deploy a Conductor, or continue to [Session Smart Conductor-managed Router Deployment](#session-smart-conductor-managed-router-deployment) to deploy a Conductor-managed router.
+
+## Selecting the Instance Size
+
+The following instance types are supported for virtual SSR in GCP. Choose the size that best meets your requirements. For additional machine type details, refer to the [GCP machine type documentation](https://docs.cloud.google.com/compute/docs/compute-optimized-machines#c2_machine_types).
+
+| Recommended VM Size | Max vNICs Supported | vCPU Cores | Memory |
```
MichaelBaj, line 30: These are the instance sizes we recommend, but for what size deployment? For conductors? For routers? I think it would be good to provide a link to our conductor scaling recommendations, if this is to be for a conductor GCP instance: intro_system_reqs.md#conductor-scaling-recommendations

docs/intro_installation_byol_gcp_mist.md
```diff
@@ -0,0 +1,238 @@
...
+**Bring Your Own License (BYOL):** BYOL allows you to install your own licensed copy of the SSR software on a GCP VM. The device registration code is used to authenticate access to the Mist installation repositories.
+
+For the latest information about SSR BYOL offerings, refer to the [Cloud Images BYOL Release Notes](release_notes_byol.md).
+
+
+## Selecting the Instance Size
+
+The following instance types are supported for virtual SSR in GCP. Choose the size that best meets your requirements. More information can be found in the [GCP Documentation](https://docs.cloud.google.com/compute/docs/compute-optimized-machines#c2_machine_types).
+
+| Recommended VM Size | Max vNICs Supported | vCPU Cores | Memory |
```
MichaelBaj, line 21: Similar to my sizing comment from before, since we don't have router sizing on our docs site, perhaps provide another column with the equivalent SSR product? (e.g., SSR120, SSR130, etc.)

MichaelBaj (changes requested):

docs/intro_installation_byol_gcp_mist.md
```diff
@@ -0,0 +1,238 @@
...
+
+1. Management (optional)
+2. WAN interfaces (required, at least one)
+3. LAN interfaces (required, at least one)
+4. HA Sync (optional, required when HA is enabled)
+5. HA Fabric (optional, required when HA is enabled)
+
+Interface names start at `ge-0-0` and increment in the same order shown above.
+
+Example: if an HA SSR is deployed with one Management interface, two WAN interfaces, and two LAN interfaces, names map as follows:
```
Chr1st0ph3rTurn3r: If an HA SSR is deployed with one Management interface, two WAN interfaces, and two LAN interfaces, the names map as follows:

Chr1st0ph3rTurn3r (changes requested):

docs/intro_installation_byol_gcp_mist.md
haberkornsam: I think this is a larger effort we need to do for all 3 clouds. @agrawalkaushik and I have talked about how we should do this.

Chr1st0ph3rTurn3r (approved):

docs/intro_installation_byol_gcp_mist.md
agrawalkaushik: @haberkornsam my understanding is that we picked these instance types based on the SSR models, correct? Can we try to let copilot solve this for us? @MichaelBaj my one hesitation of adding this level of detail is that its not apples to apples - so we might want to be mindful of the wording.

to Sam's point - we would want to tackle that in a separate PR

MichaelBaj (approved):

PR #1024, opened 2026-05-26
DOCS-540 Add clarifying language for FIPS non-compliant algorthims

MichaelBaj:


Chr1st0ph3rTurn3r (approved):

rfish-juniper (approved):

docs/_fips_non_fips_algorithms.md
```diff
@@ -0,0 +1,54 @@
+<!---FIPS default behavior in 7.0 / 7.1 and the impact on non-FIPS algorithms--->
+
+:::important
+Beginning with SSR software **version 7.0** (including **7.1**), the SSR installs with **FIPS mode enabled by default**. In these releases, FIPS enforcement prevents the use of non-FIPS-compliant cryptographic algorithms such as **MD5** and **SHA-1**. This affects features that rely on these algorithms, including:
```
agrawalkaushik, line 4: nit: `including 7.1` is implied in hte `Beginning... 7.0`. Otherwise we might have to keep this updated for 7.2 and other versions in the future.

docs/_fips_non_fips_algorithms.md
```diff
@@ -0,0 +1,54 @@
...
+
+- BGP MD5 neighbor authentication (`auth-password`).
```
agrawalkaushik, line 6: I know this is not meant to be an exhaustive list - but this often comes up in context of IPSec so would be good to add it here.

docs/_fips_non_fips_algorithms.md
```diff
@@ -0,0 +1,54 @@
...
+- MSDP peer/mesh-group authentication (`auth-password`).
+- OSPF MD5 message-digest authentication.
+- NTP client authentication using `md5`.
+- RADIUS, which uses MD5 internally to protect attributes and the shared secret.
+- Any other feature that negotiates or stores a hash using MD5 or SHA-1.
+
+If you need to use one of these features on a 7.0 or 7.1 system, FIPS mode must be disabled on the affected node **before** the feature is configured. See [Disable FIPS Mode](#disable-fips-mode-to-use-non-fips-algorithms) below.
+
+In future SSR releases, FIPS will continue to be enabled by default, but it will no longer prevent the configuration of non-FIPS algorithms. FIPS compliance will become *compliance by configuration* — operators who require strict FIPS compliance must avoid configuring non-FIPS algorithms, but the software will not block them.
```
agrawalkaushik, line 15: are we forced to add this future note for some customer commit? Just curious coz this will raise the question of which exact release.

docs/_fips_non_fips_algorithms.md
agrawalkaushik, line 6: @jamesesilvia does IDP require FIPS mode to be disabled as well?

docs/_fips_non_fips_algorithms.md
MichaelBaj: These versions are explicitly called out because I want to keep these sections updated until we change the behavior of the system to be compliance by configuration.

docs/_fips_non_fips_algorithms.md
MichaelBaj: What algorithms are excluded during IPSec key negotiation?

docs/_fips_non_fips_algorithms.md
MichaelBaj: The future release should be 7.2.3, but I believe it just got bumped again. You tell me. :)

agrawalkaushik: That's my worry - my thought was to not add any future looking language. But I'll defer to your best judgement.

PR #1112, opened 2026-08-10
IP SAN requirement for CA cert used in SCO.

Chr1st0ph3rTurn3r:


docs/sec-conductor-onboard.md
```diff
@@ -28,6 +28,10 @@ To provide a secure and mutually authenticated onboarding mechanism, the followi
 
 The public certificate and CA certificate are configured on the conductor at the Authority level.
 
+Secure Conductor Onboarding requires the use of an IP SAN (subject alternative name) for each router that is onboarding to the conductor. This requires that the certificate issued by the CA includes includes the IP SAN, which allows the conductor to perform validation and authentication.
```
MichaelBaj: Does the IP SAN need to match any fields in the configuration relating to the router? If so, that would be worth mentioning here as well, unless that is otherwise fully covered in the related link.

gregschrock: "for each router" makes it sound like the SAN is required multiple times. But it's only a requirement for the conductor's certificate (more of a one-time thing). I'd say `...for routers to securely identify the conductor's identity.` The second sentence would need a similar adjustment as the SAN is for router's authenticating the conductor, not vice verse.

I'd also call out that multiple IP SAN entries are required for HA conductors (one for each node).

gregschrock: The SAN(s) would need to match the conductor IPs that are configured in the mist redirect site. The API is `/api/v1/sites/<site>/setting` and the relevant subset of the content looks like this:
```json
{
    "ssr": {
        "conductor_hosts": ["10.92.53.7", "10.92.54.7"],
        "conductor_token": "<sco-token>"
    }
}
```
The IP SAN would need to include those `conductor_hosts` IPs as they are what the router will use to connect to the conductor.

madamsJuniper (dismissed):

MichaelBaj (dismissed):

madamsJuniper (dismissed):

gregschrock (dismissed):

docs/sec-conductor-onboard.md
gregschrock: My prior comment was specific to redirect. More generically, the relevant config would be the `authority -> conductor-address` list which specifies addresses routers should use to reach the conductor.

gregschrock (approved):
LGTM

PR #1113, opened 2026-08-11
dynamic-source-nat and IDP doc updates

jamesesilvia:
## Description
As seen in customer setup (I95-65301), with IDP + Dynamic Source NAT -- the `applies-to-local-breakout` flag must also be configured.


docs/config_dnat.md
```diff
@@ -67,6 +67,7 @@ Dynamic Source NAT is configured from the network-interface using the following
 
 - [`local-ip`](config_command_guide.md#configure-authority-router-node-device-interface-network-interface-dynamic-source-nat-local-ip): For packets ingressing this interface, the IP that is source NAT'ed to the `remote-ip`. 
 - [`remote-ip`](config_command_guide.md#configure-authority-router-node-device-interface-network-interface-dynamic-source-nat-remote-ip): For packets ingressing this interface, the IP where the `local-ip` will be source NAT'ed. `remote-ip` must use the /32 prefix.
+- [`applies-to-local-breakout`](config_command_guide.md#configure-authority-router-node-device-interface-network-interface-dynamic-source-nat-applies-to-local-breakout): Whether the dynamic source NAT applies to local breakout sessions. Default is `false`. Set to `true` when IDP is enabled on traffic that also requires Dynamic Source NAT — IDP processes traffic as local breakout sessions, so without this flag the NAT rules are not applied to that traffic.
```
MichaelBaj, line 70: Is there a case where you would set this to `true` and IDP is not configured? i.e., is this only relevant for when IDP and local-break-out is in use together?

docs/ts_idp.md
```diff
@@ -59,6 +59,18 @@ Beginning with SSR version 6.1.4-R2, traffic destined for IDP will temporarily b
 
 Additionally, services configured for IDP and their status (passing through IDP or not, and why), and when the IDP will be operational are reported in the `show idp application status` command.
 
+## IDP and Dynamic Source NAT
+
+When IDP is enabled on a service that also uses [`dynamic-source-nat`](config_dnat.md), traffic processed by IDP may not be source-NATed as expected. IDP processes traffic as local breakout sessions, and by default `dynamic-source-nat` does not apply to local breakout sessions (`applies-to-local-breakout` defaults to `false`).
+
+To resolve traffic issues in this scenario, set `applies-to-local-breakout true` on the `dynamic-source-nat` configuration for each node:
```
MichaelBaj: Replace "To resolve traffic issues in this scenario" with "To apply source NAT in scenarios where IDP is also required,"

MichaelBaj (changes requested):

docs/config_dnat.md
agrawalkaushik: Yes - this would also be useful for applying the dynamic source nat feature for local breakout scenarios. This was actually implemented for that purpose originally. Since IDP KNI's look like "local breakout" i.e. non SVR - that was the trick to get IDP working. I would argue this is a bit of a hack for IDP but the best option we have right now.

Chr1st0ph3rTurn3r (dismissed):
Approve with changes requested by Mike.

MichaelBaj (approved):

PR #1131, opened 2026-09-01
7.2.3-r2 release documentation

Chr1st0ph3rTurn3r:
Currently only beta release documentation for swift.


docs/release_notes_128t_7.2.md
```diff
@@ -71,6 +71,57 @@ An issue has been identified that may be observed in conductor deployments runni
 
 An issue has been identified when onboarding SSR routers installed with older versions of software (such as 5.4.4) to Conductors running 6.3.x, when running in offline-mode. In some cases, certain software packages are not available to be installed during onboarding. To work around this issue, import the **package-based** (the "128T" prefixed) ISO for the current conductor version onto the conductor. This provides the necessary software packages to complete the onboarding process. This issue will be resolved in a future release. 
 
+## Release 7.2.3-3r2 
+
+**Beta Release Date:** September 1, 2026
+
+### New Features
+
+- **I95-62790 ML-DSA for SVR Certificates:** ML-DSA has been added as a new algorithm option to the existing customer-facing certificate lifecycle. ML-DSA support applies exclusively to SVR peering certificates managed by the customer through the REST API. For more information, see [ML-DSA](sec_enhanced_key_mgmt.md#ml-dsa).
```
MichaelBaj: remove "customer-facing" (not sure why that was added).

docs/release_notes_128t_7.2.md
```diff
@@ -71,6 +71,57 @@ An issue has been identified that may be observed in conductor deployments runni
...
+------
+- **I95-63985 VRRP Non-revertive Active/Active Recovery:** Added support for VRRP to automatically revert from an active/active state back to active/standby when the underlying Layer 2 connectivity is restored, without requiring manual intervention such as priority changes or interface flaps.
+------
+- **I95-63281 Configurable serial console baud rate:** Added the ability to configure the serial baud rate to either 9600, 38400 or 115200 (default). This can be done from the command line `config authority router <router> node <node> serial-console-baud-rate`, or the web interface at `Router -> Node -> Serial Console Baud Rate` using the dropdown.
```
MichaelBaj: Is this restricted to a specific set of platforms or all platforms? Either way, it should be called out.

docs/release_notes_128t_7.2.md
```diff
@@ -71,6 +71,57 @@ An issue has been identified that may be observed in conductor deployments runni
...
+
+### Resolved Issues
+
+- **I95-62536 ESKM Peer Path Teardown on Certificate Validation Failure:** Resolved an issue where an ESKM router continued to accept traffic from a peer even when peer paths back to it were down, or when certificate validation was set to `fail-hard` mode. Stale peer metadata keys are now removed upon validation failure or certificate expiration to terminate invalid traffic.
```
MichaelBaj: I find the wording a bit confusing. Recommended alternative:

ESKM Peer Path Teardown on Certificate Validation Failure:** Resolved an issue where an ESKM router continued to accept traffic from a peer even its peer path is down, or when certificate validation was set to `fail-hard` mode.

docs/sec_enhanced_key_mgmt.md
```diff
@@ -300,13 +301,115 @@ The peer list of the router must also have the `peering-common-name` of that pee
...
+- ML-DSA is **not supported on devices that are in FIPS-mode**. In order to configure ML-DSA certificates, you must disable FIPS.
+
+- ML-DSA certificates must be created and provisioned using the SSR’s REST APIs and configured as file references They cannot be configured as content directly in the configuration.
+
+- Any certificate with an ML-DSA public key must be signed with an ML-DSA signature. The reverse is also true: any certificate with a classical public key (eg. RSA, ECC) must be signed with a classical signature algorithm. This also applies to intermediate CA certificates.
+
+
+#### Prerequisites
+
+In order to configure an ML-DSA certificate on a router, you must first enable post-quantum cryptography on each peer. After modifying this setting, the 128T service must be restarted for the change to take effect. Once the service has been restarted, you will be able to provision ML-DSA certificates.
```
MichaelBaj: What happens if you don't do this? What is the purpose of this configuration setting? Why isn't the certificate sufficient?

MichaelBaj (changes requested):

docs/sec_enhanced_key_mgmt.md
plessard128: If you don't do this, when you try to provision ML-DSA certificates, validation will fail and they will be rejected.

The reason for this configuration setting is FIPS-related. Since like 10 years ago, our C++ apps run with openssl set to FIPS-enabled, even if the system is not in FIPS mode at the OS level. As I understand it, this was in place as a guardrail. The openssl fips provider we use does not support ML-DSA, so in order to use ML-DSA certificates, we have to disable that guardrail.

It was decided that it would be undesirable to universally remove the guardrail for all systems, even if they aren't using PQC, so this configuration setting disables it only for customers that are actually using ML-DSA. (Also worth noting that if the OS is in FIPS mode, that takes precedence over this config setting, so it does not break FIPS compliance for customers who actually need it)

docs/sec_enhanced_key_mgmt.md
```diff
@@ -300,13 +301,114 @@ The peer list of the router must also have the `peering-common-name` of that pee
...
+exit
+```
+
+#### Configuration - Trusted CA Certificates
+
+ML-DSA Trusted CA certificates are configured and provisioned using the SSR’s REST API. Note that this procedure is different from the currently documented procedure for [installing a trusted CA certificate](config_custom_certs.md#install-the-trusted-ca-certificate). ML-DSA certificates cannot be configured as `content`; they must be ingested via the REST API.
+
+##### Ingest the Certificate
+
+When the signed certificate is returned, instruct the SSR to ingest the certificate. The certificate must be associated with the name used in the two earlier API calls. Create the following json file:
```
plessard128: This makes a reference to `When the signed certificate is returned` and `two earlier API calls`, which I think comes from the procedure to ingest client certificates, but there is nothing returning a signed certificate or earlier API calls for the workflow to ingest a trusted CA certificate. You just take the CA's certificate and you ingest it.

docs/sec_enhanced_key_mgmt.md
```diff
@@ -300,13 +301,114 @@ The peer list of the router must also have the `peering-common-name` of that pee
...
+
+**certificate.json**
+
+```
+POST /api/v1/router/{router_name}/node/{node_name}/trusted-ca-certificate
+{
+    "name": "my_peering_cert",
```
plessard128: The examples here all name the certificate `my_peering_cert`, which might be a bit confusing since this is supposed to be a CA cert, not a peering cert. Maybe `my_ca_cert`?

docs/release_notes_128t_7.2.md
```diff
@@ -71,56 +55,295 @@ An issue has been identified that may be observed in conductor deployments runni
...
+- **I95-62790 ML-DSA for SVR Certificates:** ML-DSA has been added as a new algorithm option to the existing certificate lifecycle. ML-DSA support applies exclusively to SVR peering certificates managed by the customer through the REST API. For more information, see [ML-DSA](sec_enhanced_key_mgmt.md#ml-dsa).
+------
+- **I95-63281 Configurable serial console baud rate:** Added the ability to configure the serial baud rate on the SSR devices to either 9600, 38400, or 115200 (default). This can be done from the command line `config authority router <router> node <node> serial-console-baud-rate`, or the web interface at `Router -> Node -> Serial Console Baud Rate` using the dropdown.
 
+  :::note
+  Since the baud rate change happens at the kernel level, output such as bios or grub remains at 115200bps. If the serial console client is set to a value other than 115200, the output will be displayed as garbage. If one of those garbage characters happens to be `xon`, the client will stop the display. To prevent the display issue, it is recommended to turn off software flow control (`xon/xoff`).
```
MichaelBaj: replace "garbage" with "garbled text"

docs/release_notes_128t_7.2.md
```diff
@@ -71,56 +55,295 @@ An issue has been identified that may be observed in conductor deployments runni
...
 ------
-- **I95-63985 VRRP Non-revertive Active/Active Recovery:** Added support for VRRP to automatically revert from an active/active state back to active/standby when the underlying Layer 2 connectivity is restored, without requiring manual intervention such as priority changes or interface flaps.
+- **I95-64692 Improved L7-Security Packet Interface:** The internal packet interface used for L7-Security function chaining, such as IDP, has moved from the deprecated KNI (Kernel NIC Interface) technology to virtio-user. This improves throughput headroom for IDP-enabled deployments and removes a dependency that upstream DPDK is phasing out.
```
MichaelBaj: Great explanation, just not one that I would frame for customers in this way.

- **I95-64692 Improved L7-Security Packet Performance:** The internal packet interface used for L7-Security function chaining, such as IDP, has changed affording more throughput headroom for IDP-enabled deployments.

docs/release_notes_128t_7.2.md
```diff
@@ -71,56 +55,295 @@ An issue has been identified that may be observed in conductor deployments runni
...
+------
+- **The following issues have been addressed and delivered to increase the overall security posture of the SSR:** I95-62091, I95-65017, I95-65018, I95-65021, I95-65025, I95-65026, I95-65027, I95-65028, I95-65030, I95-65033, I95-65034, I95-65038, I95-65039, I95-65044, I95-65054, I95-65055, I95-65205, I95-65206, I95-65208, I95-65210, I95-65211, I95-65217, I95-65219, I95-65221, I95-65224, I95-65226, I95-65228, I95-65235, I95-65238, I95-65239, I95-65247, I95-65249, I95-65250, I95-65252, I95-65254, I95-65297.
+------
+- **I95-58472 SSR4x0 Platform Provisioning Reboot Loop:** Resolved an issue where CN102 SSR4x0 platforms could enter an infinite reboot loop during initial provisioning.
```
MichaelBaj: - **I95-58472 SSR4x0 Platform Provisioning Reboot Loop:** Resolved an issue where SSR4x0 platforms could enter an infinite reboot loop during initial provisioning.

docs/release_notes_128t_7.2.md
```diff
@@ -71,56 +55,295 @@ An issue has been identified that may be observed in conductor deployments runni
...
+------
+- **I95-62758 BGP Establishment Taking Incorrect Path Over GRE:** Resolved an issue where BGP SYN packets were sent over the wrong GRE tunnel instead of using the direct path, causing BGP to establish over an unintended interface.
+------
+- **I95-63195 Highway Crash During ESKM Session Scaling:** Resolved an issue where the highway process could crash during high session scaling due to incorrect metadata propagation in ESKM encrypt/decrypt contexts.
+------
+- **I95-63811 IPv6 Service with ICMP Transport Not Routed:** Resolved an issue where PINGv6 sessions were not routed when an IPv6 service was configured with ICMP as the transport protocol. The system now correctly remaps ICMP to ICMPv6 for IPv6 service prefixes.
+------
+- **I95-63876 Route Flapping and Inaccessibility:** Resolved an issue where routes would flap or become inaccessible in hub-and-spoke topologies with inter-hub steering preferences configured, causing intermittent connectivity failures.
```
MichaelBaj: - **I95-63876 BGP Route Flapping and Inaccessibility:** Resolved an issue where BGP routes would flap or become inaccessible in hub-and-spoke topologies with inter-hub steering preferences configured, causing intermittent connectivity failures.

docs/release_notes_128t_7.2.md
```diff
@@ -71,56 +55,295 @@ An issue has been identified that may be observed in conductor deployments runni
...
+------
+- **I95-63895 SSR sending packets larger than configured MTU:** Resolved an issue where the SSR was sending packets larger than the configured MTU (e.g., 1518 bytes instead of 1500), causing packet drops on downstream network elements.
+------
+- **I95-63913 Session-source incorrect in BFD pinhole:** Resolved an issue where session-source was incorrectly set to public when a BFD pinhole also happened to be a flow-move scenario.
+------
+- **I95-64054 Highway Crash with ESKM Jumbo Packets:** Resolved an issue where the highway process could crash with a null pointer dereference when processing jumbo packets with ESKM due to an uninitialized encryption context.
```
MichaelBaj: Please do not offer details into why the system would crash. This offers a bad actor a window into exploiting the product if they know a customer is running a prior version.

- **I95-64054 Highway Crash with ESKM Jumbo Packets:** Resolved an issue where the highway process could crash when processing jumbo packets with ESKM due to an uninitialized encryption context.

docs/release_notes_128t_7.2.md
```diff
@@ -71,56 +55,295 @@ An issue has been identified that may be observed in conductor deployments runni
...
+------
+- **I95-64603 `Chronyd` Requires Manual Restart After Reboot:** Resolved an issue where all NTP servers appeared as rejected after a reboot, requiring a manual restart of the `chronyd` service to restore time synchronization.
+------
+- **I95-64610 Peer Name in `show waypoints`:** The `show waypoints` command output now includes a peer name column, making it easier to identify which peer is associated with a given waypoint entry.
+------
+- **I95-64627 Certificate Unavailable for Peering After Upgrade:** Resolved an issue where the local certificate became unavailable for peering after an upgrade, resulting in peer paths remaining down with a "No local certificate available" error.
+------
+- **I95-64684 HMAC Cipher Mode Information in Logs and Session Output:** Added HMAC mode and HMAC cipher details to session logs and `show sessions` output, improving visibility into the active encryption and authentication state of sessions.
+------
+- **I95-64688 Highway coredumps causing peer path flaps:** Resolved an issue where highway process coredumps were occurring, resulting in peer path flaps.
```
MichaelBaj: This needs an explanation as to the trigger. Please work with the developer to get this information.

docs/release_notes_128t_7.2.md
```diff
@@ -71,56 +55,295 @@ An issue has been identified that may be observed in conductor deployments runni
...
+------
+- **I95-64811 Highway crash causing session drops:** Resolved a highway process crash that occurred under specific traffic conditions, resulting in session drops and temporary traffic disruption.
```
MichaelBaj: Any more clarity on the "specific traffic conditions"?

docs/release_notes_128t_7.2.md
```diff
@@ -71,56 +55,295 @@ An issue has been identified that may be observed in conductor deployments runni
...
+------
+- **I95-64903 High CPU and Disk Usage on SSR440:** Resolved an issue where SSR440 devices experienced high CPU and disk usage without corresponding syslog messages being generated, making the condition difficult to diagnose.
+------
+- **I95-64905 401 Authorization Required error when refreshing Logs page:** Resolved an issue where refreshing the Logs page on the conductor GUI returned a 401 Authorization Required error, requiring a full page reload or re-login.
+------
+- **I95-64908 Runtime Error Collapsing Logs Panel in Web Interface:** Resolved an issue where interacting with the collapse control on the Conductor Logs page before the page had fully rendered could trigger a runtime script error. The Logs page now safely handles this interaction.
+------
+- **I95-64978 Highway crash on head-end router causing interface flaps:** Resolved an issue where a race condition on session classification fields (such as domain name, URI, and application classification) could cause the highway process to crash with a core dump, resulting in interface flaps. Access to these fields is now properly synchronized.
```
MichaelBaj: "... could cause the highway process to crash, resulting in interface flaps."

docs/release_notes_128t_7.2.md
```diff
@@ -71,56 +55,295 @@ An issue has been identified that may be observed in conductor deployments runni
...
+------
+- **I95-65314 Sessions Not Switching to Available Source NAT IPs:** Resolved an issue where a source NAT database corruption and race condition prevented sessions from switching to additional available IP addresses on a WAN interface, causing packet drops.
+------
+- **I95-65336 Factory reset resilience to interruption:** Improved the factory reset procedure to be more resilient to interruption (e.g., unexpected reboot). The system now tracks reset progress and can resume or indicate completion status after recovery.
+------
+- **I95-65337 Missing FIB Entries After Router Migration:** Resolved an issue where FIB entries were missing after migrating WAN interfaces from one router to another using the same IP addresses. The system now correctly detects new peers and triggers path addition for peers that are already up.
+------
+- **I95-65348 Added Support for Bouncing PoE Ports:** Added the ability to force a link down/up cycle ("bounce") on SSR4x0 PoE ports, allowing a connected access point to be power-cycled without physically disconnecting the cable.
```
MichaelBaj: This should be moved to the new feature section

docs/release_notes_128t_7.2.md
```diff
@@ -71,56 +55,295 @@ An issue has been identified that may be observed in conductor deployments runni
...
+------
+- **I95-65351 IMA and Security Stack Incompatibility:** Resolved an issue where enabling IDP with IMA caused the security engine to fail to start due to an incompatibility between IMA enforcement and the security stack binaries.
+------
+- **I95-65354 Missing Dependencies in Offline ISO:** Resolved an issue where required software dependencies were missing from the offline ISO, preventing plugin upgrades in air-gapped environments.
+------
+- **I95-65365 PCLI Command to Trigger GARP:** Added the PCLI command to manually trigger Gratuitous ARP (GARP) on VRRP interfaces, accepting device and network-interface as arguments.
```
MichaelBaj: move to new feature section?

docs/release_notes_128t_7.2.md
```diff
@@ -71,56 +55,295 @@ An issue has been identified that may be observed in conductor deployments runni
...
+------
+- **I95-65366 Maximum GARP interval for VRRP:** Added a configurable `maximum-garp-interval` parameter for VRRP, allowing control over how frequently gratuitous ARP messages are sent during VRRP state transitions. This prevents excessive ARP traffic in environments with many VRRP instances.
```
MichaelBaj: move to new feature section?

docs/release_notes_128t_7.2.md
```diff
@@ -71,56 +55,295 @@ An issue has been identified that may be observed in conductor deployments runni
...
+------
+- **I95-65478 Mist-Agent Upgrade Blocked by Version Mismatch:** Resolved an issue where a router running a newer `mist-agent` package version than the one bundled with the target software release could fail to upgrade, because `mist-agent` downgrades were not permitted by default.
+------
+- **I95-65486 Highway Crash During Upgrade from Legacy Version:** Resolved a highway crash that occurred during router upgrades from legacy versions (e.g., 5.5.x to 7.2.x), causing the upgrade to fail after timing out.
+------
+- **I95-65488 Improved Logging for File-Based Trusted CA Certificates:** Added logging to the configuration director to make it easier to determine when a file-pointer-based trusted CA certificate is referenced in configuration but the file is not yet present on disk.
+------
+- **I95-65512 Web Server and Nginx Crash Looping:** Resolved an issue where the web server and nginx processes entered a crash loop, preventing access to the SSR Web UI.
```
MichaelBaj: Please work with the developer to better understand the condition triggering this crash loop.

docs/release_notes_128t_7.2.md
```diff
@@ -71,56 +55,295 @@ An issue has been identified that may be observed in conductor deployments runni
...
+------
 - **I95-65544 Host Service Exposure on WAN Interface During Startup:** Resolved an issue where host services (such as SSH) were exposed on external WAN interfaces during system boot because `firewalld` flushed interface-to-zone associations during startup reloads before security rules were fully established.
 ------
+- **I95-65545 Incorrect Fragmentation Stats Table Name:** Resolved an issue where the PCLI displayed an incorrect table name ("Non-Fabric IPv6 Fragmented Packets" instead of "Non-Fabric IPv4 Fragmented Packets") for IPv4 fragmentation statistics.
+------
+- **I95-65548 DSCP steering support with deferred classification in hierarchical services:** Added support for DSCP steering services when classification is deferred in hierarchical service configurations, enabling correct traffic handling in Mist-managed deployments.
+------
+- **I95-65557 Highway Crash During Show Commands on HA Router:** Resolved an issue where issuing show commands on an HA router could cause a highway crash on both nodes due to FIB table contention. FIB table operations are now batched to prevent mutex lock errors.
```
MichaelBaj: arguably this should only be `show fib` commands, not all commands.

docs/release_notes_128t_7.2.md
```diff
@@ -71,56 +55,295 @@ An issue has been identified that may be observed in conductor deployments runni
...
 ------
 - **I95-65603 Peer Path UP with Mixed Certificates in Fail-Hard Mode:** Resolved an issue where peer paths remained UP between nodes using mismatched certificate trust anchors (default factory-signed and custom CA-signed certificates) even when `invalid-certificate-behavior` was configured to `fail-hard`.
 ------
 - **I95-65609 Routers Reverting to Waiting State After Conductor Upgrade:** Resolved an issue where previously onboarded routers incorrectly reverted to a "waiting" state in `show secure-conductor-onboarding` following a conductor upgrade due to improper handling of startup HTTP responses during authorized key retrieval.
 ------
+- **I95-65617 Loss of syslog forwarding over TLS after upgrade:** Resolved an issue where syslog forwarding over TLS stopped working after upgrading, preventing log delivery to remote collectors.
+------
+- **I95-65635 Source NAT Port Exhaustion on Loopback Interface:** Resolved an issue where a large number of `SourceNatPortException` errors for the local KNI interface caused SSH connection failures to the SSR loopback IP. Host-type service routes no longer use the KNI IPv6 control interface for source NAT.
```
MichaelBaj: A customer won't understand how this relates to them. The loopback interface on the product is an internal detail. This needs to be translated to customer experience.

docs/release_notes_128t_7.2.md
```diff
@@ -71,56 +55,295 @@ An issue has been identified that may be observed in conductor deployments runni
...
 - **I95-65669 Configuration Synchronization Following Conductor Migration:** Resolved an issue where a router appeared synchronized after conductor migration but retained stale configuration because a temporary `NONE_AVAILABLE` response from the conductor was treated as terminal, halting configuration retrieval until a manual commit was executed.
 ------
-- **I95-65855 IPv6 Link-Local Resolver Nginx Failures:** Resolved an issue where Nginx failed to start or entered a crash loop when no external DNS was configured because IPv6 link-local addresses were selected as resolvers. The resolver logic now filters link-local addresses, properly formats IPv6 addresses, and prioritizes `/etc/hosts` and `dnsmasq` overrides.
+- **I95-65680 RoutingManager Not Running on HA Headend Router:** Resolved an issue where the routingManager could remain in STANDBY after a session interruption, leaving the router without an active routing process (loss of BGP/routing connectivity) until restarted.
+------
+- **I95-65691 Node disconnected after headend partial rollback:** Resolved an issue where a node could remain disconnected from the conductor after a partial rollback scenario on a headend router.
```
MichaelBaj: This would benefit from a better explanation for a "partial rollback"

docs/release_notes_128t_7.2.md
```diff
@@ -71,56 +55,295 @@ An issue has been identified that may be observed in conductor deployments runni
...
+------
+- **I95-66067 Offline upgrade failure:** Resolved an issue during upgrade that was being reported as an `Unpacker Failure`. The service startup order has been adjusted to prevent the issue in future upgrades/installations.
+------
+- **I95-66071 Resolved a Highway Crash During GRE Tunnel Configuration Updates:** Resolved an issue where a failed lookup during a GRE tunnel interface modification could leave stale internal state, causing the highway process to crash on a subsequent configuration change to the same tunnel. GRE tunnel state is now cleaned up correctly when a lookup fails.
+------
+- **I95-66077 BGP Peers Down After Enabling a Second Internet Provider:** Resolved a double-free defect in BGP conditional advertisement processing that could cause all BGP peers to go down when a second BGP-connected internet provider was enabled.
```
MichaelBaj: - **I95-66077 BGP Peers Down After Enabling a Second Internet Provider:** Resolved a defect in BGP conditional advertisement processing that could cause all BGP peers to go down when a second BGP-connected internet provider was enabled.

MichaelBaj (changes requested):

PR #1142, opened 2026-09-09
LAG/LACP is not supported on the SSR4x0 platforms

Chr1st0ph3rTurn3r:


rfish-juniper (approved):
LGTM, thanks!  Are there any other references in the doc set that will need to be addressed?

MichaelBaj (approved):

dmontgomeryJuniper (approved):

PR #1160, opened 2026-09-23
6.2.15 release notes

Chr1st0ph3rTurn3r:


docs/release_notes_128t_6.2.md
```diff
@@ -24,6 +24,19 @@ Before upgrading please review the [**Upgrade Considerations**](intro_upgrade_co
 ------
 - **Plugin Upgrades:** If you are running with plugins, updates are required for some plugins **before** upgrading the conductor to SSR version 5.4.0 or higher. Please review the [Plugin Configuration Generation Changes](intro_upgrade_considerations.md#plugin-configuration-generation-changes) for additional information.
 
+## Release 6.2.15-3-lts
+
+**Release Date:** September 24, 2026
+
+### Resolved Issues
+
+- **I95-66234 Creating a snapshot log for large configurations appears to hang the system:**
```
MichaelBaj: - **I95-66234 Creating a snapshot log for large configurations appears to stall the system:** 

Stall is better than hang since it implies that it will come back after the process completes, which is true.

MichaelBaj (changes requested):

MichaelBaj (approved):

